PR #10, opened 2017-03-24
Speedups to check_islands_for_overlap

mhardcastle:
I sent these changes to David a couple of months ago and they are circulating in the surveys KSP, but a formal pull request seems like much the best way of dealing with them. (Well done for getting this on Github, guys!)

Rationale: when ``check_islands_for_overlap()`` (``wavelet_atrous.py``) runs on a large image (e.g. the 20000 x 20000 images we are generating in the Tier 1 pipeline) it can take a very long time.Typical run times are several days for multiple calls, e.g. with ``jmax = 4``. 

The basic issue is that ``check_islands_for_overlap`` is doing several expensive operations on numpy arrays the size of the whole image -- essentially ``wav_islands`` is an array the size of the whole image
containing wavelet island numbers where they overlapped with an existing Gaussian. But we're doing that over all of ``wimg.islands``; it's much cheaper to do it only for the island we care about at that point in the loop, using the bounding box of that island only. I think the lines

```
wav_islands = orig_rankim_bool[wvisl.bbox] * wpp[wvisl.bbox] - 1
if idx in wav_islands:
    ....
```
basically will catch all cases where the wavelet gaussian idx overlaps with an existing source. (No need to uniquify this because we only use ``wav_islands`` for this one operation and it's quicker just to search for the value we want.) Then similarly the operation of finding the indices of the overlapping sources can be done in the bounding box only.

``merge_islands`` also did expensive operations over the whole field and I've again restricted it to working over the joint bounding box of the two islands. A change to the ``Island`` class would be needed to avoid the bit where I generate the full-size labels image, which would be worthwhile.

The overall effect is to speed up the runs over Tier 1 from a few days to a few hours, though checking for overlap is still the slowest operation. But it would be great if you could have a look over this to see if you agree that these are sensible changes.

### DIFF
--- a/bdsf/wavelet_atrous.py
+++ b/bdsf/wavelet_atrous.py
@@ -571,6 +571,30 @@ def fftconvolve(in1, in2, mode="full", pad_to_power_of_two=True, numcores=1):
     elif mode == "valid":
         return _centered(ret, abs(s2 - s1) + 1)
 
+
+def rebase_bbox(box,minv):
+    # return a new bounding box tuple where minv is subtracted from
+    # all the co-ordinate values
+    nbox=[]
+    for i,sl in enumerate(box):
+        nbox.append(slice(sl.start-minv[i],sl.stop-minv[i],None))
+    return tuple(nbox)
+
+def merge_bbox(box1,box2):
+    # For two bounding box tuples find the minimal n-dimensional space
+    # that encompasses both structures and make new bounding boxes in
+    # this co-ordinate system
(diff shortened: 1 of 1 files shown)


darafferty:
Thanks, Martin. I opened an issue of the LOFAR issue tracker (https://support.astron.nl/lofar_issuetracker/issues/10433) for this a few weeks ago and had started implementing your changes, but a pull request makes it much easier.

PR #43, opened 2017-11-21
Using external noise and mean maps

cyriltasse:
Ok - I've tested that mode that works well, and allows to save a lot of time when the noise map has been made once. The option `rmsmean_map_filename` appears in advanced options. As suggested by @darafferty it's now using `read_image_from_file` instead of pyfits.


darafferty:
Thanks -- looks good!

PR #51, opened 2018-02-22
change how setup.py builds the static libs

sabourke:
Adds a separate command for building the static fortran libs and call it as part of "build_ext", does something similar for "clean". This addresses lofar-astron/PyBDSF/issues/45


tammojan:
Great, thanks. Could you enlighten me on the `mbuild` and `mclean` commands? Is it from numpy? (Googling gives me programmers called McLean...)

PR #55, opened 2018-04-10
Add support for Boost 1.65 and above

zhengmeyer:
Add support for Boost 1.65 and above, because boost/python/numeric.hpp has een removed from Boost version 1.65 and above.

### DIFF
--- a/src/c++/num_util/num_util.cpp
+++ b/src/c++/num_util/num_util.cpp
@@ -148,7 +148,7 @@ KindCharMapEntry kindCharMapEntries[] =
     KindCharMapEntry(PyArray_CDOUBLE,'D'),
     KindCharMapEntry(PyArray_OBJECT, 'O')
   };
-  
+ 
 typedef KindTypeMap::value_type  KindTypeMapEntry;
 KindTypeMapEntry kindTypeMapEntries[] =
   {
@@ -164,14 +164,10 @@ KindTypeMapEntry kindTypeMapEntries[] =
     KindTypeMapEntry('O',PyArray_OBJECT)
   };
 
-  
 int numStringEntries = sizeof(kindStringMapEntries)/sizeof(KindStringMapEntry);
 int numCharEntries = sizeof(kindCharMapEntries)/sizeof(KindCharMapEntry);
 int numTypeEntries = sizeof(kindTypeMapEntries)/sizeof(KindTypeMapEntry);
 
-
-using namespace boost::python;
-  
 static KindStringMap kindstrings(kindStringMapEntries,
                                    kindStringMapEntries + numStringEntries);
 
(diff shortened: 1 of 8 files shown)


tammojan:
A couple of things:
 * The part that defines `pyndarray` is duplicated 3 times. Perhaps you can factor it out into one header?
 * Travis fails, apparently `typedef typename pyndarray pyndarray` is not valid. I think you can just skip this line (the error is below).

The travis error is:
```
src/c++/MGFunction.h:11:20: error: expected nested-name-specifier before ‘pyndarray’
   typedef typename pyndarray pyndarray;
                    ^
src/c++/MGFunction.h:11:30: error: expected initializer before ‘pyndarray’
   typedef typename pyndarray pyndarray;
```

zhengmeyer:
Combined duplicated pyndarray definition into one header file. The error of `typedef typename pyndarray pyndarray` is also fixed.

src/c++/num_util/num_util.cpp
```diff
@@ -359,18 +355,24 @@ void copy_data(boost::python::numeric::array arr, char* new_data){
...
   object obj(handle<>(PyArray_NewCopy((PyArrayObject*)arr.ptr(),PyArray_CORDER)));
   return makeNum(obj);
 }
 
-  
+
 //Return a clone of this array with a new type
-numeric::array astype(boost::python::numeric::array arr, PyArray_TYPES t){
-  return (numeric::array) arr.astype(type2char(t));
+#if BOOST_VERSION < 106500
```
tammojan: It's probably cleaner to move the `ifdef` into the function definition (which is the same for both cases).

src/c++/num_util/num_util.cpp
```diff
@@ -409,6 +411,35 @@ char type2char(PyArray_TYPES t_type){
...
+    case 'f':
+      return np::dtype::get_builtin<float>();
+    case 'd':
+      return np::dtype::get_builtin<double>();
+    case 'F':
+      return np::dtype::get_builtin<std::complex<float>>();
+    case 'D':
+      return np::dtype::get_builtin<std::complex<double>>();
+    default:
+      std::cout << "nvalid character code!" << std::endl;
```
tammojan: Typo, `nvalid` is invalid.

src/c++/num_util/num_util.h
```diff
@@ -103,15 +103,15 @@ namespace num_util{
    *@param arr a Boost/Python numeric array.
    *@return a numeric array referencing the input array.
    */
-  boost::python::numeric::array makeNum(const 
-					boost::python::numeric::array& arr);
+  pyndarray makeNum(const
```
tammojan: Sorry for nagging, this can go onto one line.

src/c++/pyndarray.h
```diff
@@ -0,0 +1,8 @@
+#if BOOST_VERSION < 106500
+  typedef typename boost::python::numeric::array pyndarray;
+  //namespace np = boost::python::numeric;
+#else
+  #include <boost/python/numpy.hpp>
+  typedef typename boost::python::numpy::ndarray pyndarray;
+  namespace np = boost::python::numpy;
```
tammojan: This namespace seems to be used only in two places, perhaps better to just keep it explicit there?

zhengmeyer:
Removed namespace np, moved `ifdef` into the function. It is indeed cleaner. Typo and space are also fixed.

tammojan:
Great, thanks!!

Issue #63, opened 2018-06-12
can't process valid image


gijzelaerr:
am i missing something? following tutorial on the website but no luck:
```
BDSF [24]: filename = 'test/kat7_real_multiply45/fits/0-outputs.fits'
      ...: 

BDSF [25]: go()
--> Opened 'test/kat7_real_multiply45/fits/0-outputs.fits'
Image size .............................. : (256, 256) pixels
Number of channels ...................... : 1
Number of Stokes parameters ............. : 1
Beam shape (major, minor, pos angle) .... : (3.96088e-02, 2.73473e-02, 130.2) degrees
Frequency of image ...................... : 1825.711 MHz
Number of blank pixels .................. : 0 (0.0%)

BDSF [26]: show_fit()
Image has not been processed. Please run process_image first.

BDSF [27]: process_image()
--> Opened 'test/kat7_real_multiply45/fits/0-outputs.fits'
Image size .............................. : (256, 256) pixels
Number of channels ...................... : 1
Number of Stokes parameters ............. : 1
Beam shape (major, minor, pos angle) .... : (3.96088e-02, 2.73473e-02, 130.2) degrees
Frequency of image ...................... : 1825.711 MHz
Number of blank pixels .................. : 0 (0.0%)

BDSF [28]: show_fit()
Image has not been processed. Please run process_image first.
```

gijzelaerr:
[0-outputs.zip](https://github.com/lofar-astron/PyBDSF/files/2094778/0-outputs.zip)

darafferty:
Hmm, it seems the problem is due to an issue with the logger. I think I've fixed it in the latest master, so please try updating and running again.

gijzelaerr:
I now have output, but it looks like pybsdm is hanging on the gaussian fitting, CPU usage is 0 for the last 30 minutes:
```
BDSF [3]: go()
--> Opened '/scratch/datasets/kat7_real/raw/0-wsclean-dirty.fits'
Image size .............................. : (256, 256) pixels
Number of channels ...................... : 1
Number of Stokes parameters ............. : 1
Beam shape (major, minor, pos angle) .... : (3.96088e-02, 2.73473e-02, 130.2) degrees
Frequency of image ...................... : 1825.711 MHz
Number of blank pixels .................. : 0 (0.0%)
Flux from sum of (non-blank) pixels ..... : -0.012 Jy
--> Calculating background rms and mean images
[...]

PR #78, opened 2018-10-04
Initialize numpy to avoid segfault in Gaussian fitting (issue #63)

darafferty:
The hanging described in issue #63 seems to be caused by a segfault due to numpy not being initialized. This PR fixes it on my test system.


cyriltasse:
hey @darafferty this fixes the issue - thanks!

darafferty:
Thanks for the testing!

PR #85, opened 2018-12-29
Installation without using conda (Readme update)

kiryteo:
I faced installation issues for PyBDSF on Ubuntu 16.04 in environment without conda. After some lookup, I realized the errors being due to compiler version difference for gcc, g++ (6.5 both) and gfortran(5.5). After using update-alternatives and changing gcc, g++ version to 5.5 (same as gfortran), the installation worked without any issue. I've created a gist for 'update-alternatives' usage [here](https://gist.github.com/kiryteo/e7cb52ecce8c2b25eca6acdd8ddbb5a7).


darafferty:
Thanks for this -- hints on installing are most welcome!

Issue #100, opened 2019-06-06
error when blank_limit is set


wllwen007:
Noticed by a colleague of mine in v 1.9.0: if blank_limit is set (i.e. not None), it produces an error -- undefined `check_low`:

```        
In [4]: img = bdsf.process_image('obs1_K.fits',blank_limit=1e-10,rms_box=(50,10),thresh_isl=3.0,thresh_pix=5.0,advanced_opts=True)
--> Opened 'obs1_K.fits'
Image size .............................. : (400, 400) pixels
Number of channels ...................... : 1
Number of Stokes parameters ............. : 1
Beam shape (major, minor, pos angle) .... : (1.86122e-03, 9.54683e-04, -67.9) degrees
Frequency of image ...................... : 20061.891 MHz
Number of blank pixels .................. : 41307 (25.8%)
--> Blanking pixels with values below 1.0e-10 Jy/beam

...
...
...
    143           sys.stdout = original_stdout  # turn STDOUT back on
    144           mask_low = (count >= 5)
--> 145           if check_low:
    146               nlow = len(N.where(mask_low)[0])
    147               if nlow / float(image.shape[0] * image.shape[1]) > 0.2:

NameError: global name 'check_low' is not defined
```

Comes from here

https://github.com/lofar-astron/PyBDSF/blob/286460ecadadc24c570bc9d5d462f5727099abc6/bdsf/collapse.py#L145

where I think commit 95792ed08831c71147ee0aed4c9662935166fd65  didn't completely remove the previous references to check_low...

PR #101, opened 2019-06-06
fix blank_limit check_low error

wllwen007:
fixes https://github.com/lofar-astron/PyBDSF/issues/100


darafferty:
Looks good -- thanks!

PR #121, opened 2020-03-06
Fix boost include dirs

tammojan:
If boost is installed in a non-standard location, but boost headers are available in a system dir (e.g. `/usr/include`) building may work, but at runtime a symbol may not be found. This commit guesses the include directory from the found boost library, and passes that include directory to the build process.

We should really move away from boost_python...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -121,7 +121,7 @@ def find_library_file(libname):
 
 
 def find_boost_python():
-    """Find the name of the boost-python library. Returns None if none is found."""
+    """Find the name of the boost-python library, and the path where it was found. Returns a guess if none found."""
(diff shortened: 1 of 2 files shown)


tammojan:
@darafferty Could you have a look at this? Somehow the python3.5 test now segfaults (the button still shows green because it's an 'allowed failure').

darafferty:
I got the python 3.5 test to run successfully, but I had to retry it a few times (setting ncores=2 in my last commit might have helped, not sure).

tammojan:
@darafferty If you agree with the change, could you merge this? I think the segfault is unrelated to the change.

darafferty:
Yep, agreed.

Issue #126, opened 2020-04-06
Spectral index mode, blanking, and detection image


mhardcastle:
In spectral index mode, even if you specify a detection image, the code seems to form and use the ch0 image to find islands. (I was trying this with a beam-blanked detection image: sources outside the blanked area are found.)

Related to this, the spectral index code appears not to be able to deal with nan-blanked images, even if the detection image is blanked. (Again this may be a result of the use of the ch0 image.)

Fundamentally I think one wants to be able to supply one's own ch0 image (which may benefit from better weighting/uv coverage than the image made just by averaging the channels). However I realise I may be the only person who uses this code...

PR #127, opened 2020-04-06
Spectral index mode: fix bug and add functionality

### LINKED ISSUE #39: incl_chan=True is ignored in writing FITS
There's a long-standing bug whereby FITS catalogues don't contain channel fluxes derived in spectral index mode even if you ask for them. ASCII catalogues are fine.

I'll see if I can find time to have a look at this. My memory is that the whole FITS writing thing might benefit from a redo with astropy.tables...

mhardcastle:
Fixes #39 and adds the functionality requested in #126.

### DIFF
--- a/bdsf/collapse.py
+++ b/bdsf/collapse.py
@@ -33,7 +33,7 @@ class Op_collapse(Op):
         c_list = img.opts.collapse_av
         c_wts = img.opts.collapse_wt
         if c_list == []: c_list = N.arange(img.shape[1])
-        if len(c_list) == 1:
+        if len(c_list) == 1 and c_mode=='average':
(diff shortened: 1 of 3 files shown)


tammojan:
Hi Martin, this looks good to me. Could you perhaps document a bit clearer what the file needs to be like? I guess a FITS file with one 2d image in it?  Do the dimensions of this image need to match those of the image being processed?

mhardcastle:
Yeah, good point. Documentation of the image type added in line with that for `detection_image`.

mhardcastle:
(I should really implement this for the other stokes parameters but I don't know if anyone needs that functionality -- I certainly don't.)

PR #130, opened 2020-06-09
add boost_python3 and numpy_python3 to libnames in find functions

informationcake:
solves an issue of PyBDSF not being able to find boost library when using Python3, since boost_python can have a few name variations and these are hard coded into the setup.py file.


darafferty:
Looks good -- thanks!

PR #133, opened 2020-07-10
Use system specific multiarch name instead of x86_64

olebole:
The directory `/usr/lib/x86_64-linux-gnu` is obviously very architecture specific. However, `sys.implementation` offers a corresponding variable `_multiarch`, which is used here instead. See also https://wiki.debian.org/Python/MultiArch 
After applying this, PyBDSF [compiles on all out platforms](https://buildd.debian.org/status/logs.php?pkg=pybdsf&ver=1.9.2-2&suite=sid) (MIPS still pending):


darafferty:
Very nice -- thanks!

PR #145, opened 2021-01-05
has_pyfftw used, but not defined in wavelet_atrous.py

brentmaas:
wavelet_atrous.py appears to use a flag 'has_pyfftw' to check if the module pyfftw.interfaces could be imported, but this flag is never set, leading to a NameError as follows: https://pastebin.com/Wz7nwnH3.

Setting it True/False in the try/except block where pyfftw.interfaces is attempted to be imported should address it.

Python: 3.8.6
PyBDSF: 1.9.2


darafferty:
Yep, good catch. Thanks for the fix!

PR #150, opened 2021-03-15
Fixes #148

### LINKED ISSUE #148: PEP 517 install time dependencies
At the moment bdsf tries to import numpy as a setup.py import. This needs to be specified
as a compile time dependency to for pip to install prior to running setup.py (see PEP 517)

bennahugo:
As discussed. Makes build system compliant with PEP 517


tammojan:
Thanks!

PR #160, opened 2021-05-01
Bail out of fitting overly complex large islands, and other fixes

### LINKED ISSUE #158: Hanging for very large island
We use PyBDSF to help make the mask in the early stages of ddf-pipeline. This has been fine for normal extragalactic fields, but for Galactic fields (and possibly with the rms maps provided by Cyril's code as part of the problem) it is possible for very large islands to appear. In `gausfit.py`, what happens with these in detail is that `fit_island_iteratively` gets called, it fails and calls `fit_island` in 'default' mode, `inigaus_fbdsm` is called and gives a very large list of Gaussians (over 400 for my test case) and then when `fit_iter` is run it passes all those Gaussians to the LM minimization routine which never returns.

I think it's obviously nuts in this case to try to fit with all the free parameters of the hundreds of Gaussians -- LM can't do this even if we wanted it to -- and my 'fix' is to check if the list of Gaussians provided by `inigaus_fbdsm` is longer than some number (I have used 25) and to fall back on 'simple' fitting if so. If someone would like to comment on this strategy I can make a PR.

mhardcastle:
Fixes #158 , also a typo fix, a change to the handling of ctrl-C in non-interactive mode, and some debugging statements in verbose mode that shouldn't affect anyone -- I thought I might as well leave these in as they may be needed another time and they help to diagnose the situation if there are other bugs that cause the code to hang during a fit.

### DIFF
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -1159,7 +1159,7 @@ def read_image_from_file(filename, img, indir, quiet=False):
         reason = 0
         try:
             if not old_pyfits:
-                fits = pyfits.open(image_file, mode="readonly", ignore_missing_end=True, memmap=False)
+                fits = pyfits.open(image_file, mode="readonly", ignore_missing_end=True, memmap=True)
             else:
                 fits = pyfits.open(image_file, mode="readonly")
             img.use_io = 'fits'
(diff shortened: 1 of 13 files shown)


darafferty:
Looks great -- thanks!

PR #165, opened 2021-11-09
Improve handling of beam in spectral index module

darafferty:
This PR improves the handling of the beam when the beam_spectrum option is set to None by searching the header for the beam sizes or by scaling the size with frequency. Previously, no scaling was done.


gmloose (approved):
Looks good to me.

Issue #168, opened 2022-01-25
AttributeError: 'Extension' object has no attribute 'extra_cxx_compile_args'


MichelleLochner:
Just in case anyone has the same problem as me (or perhaps there's a better way of solving it), I got this error buried in amongst a very large traceback:

`AttributeError: 'Extension' object has no attribute 'extra_cxx_compile_args'`

I noticed that there were some very similar attributes that were initialised to empty lists in the `setup.py` file so I adjusted the for loop: 

```
for ext in extensions:
    ext.extra_f77_compile_args = []
    ext.extra_f90_compile_args = []
    ext.extra_c_compile_args = []
    ext.extra_cxx_compile_args = []
```

And everything seems to have installed ok and is running. 

I'm using Ubuntu 21.10, python 3.9 and virtualenv.

gmloose:
Thanks for reporting. This is a known issue, and I'm currently working on it. The fix for this issue is to add `Extension` to `from numpy.distutils.core import ...` in line 7.

PR #169, opened 2022-01-25
Fix py38 build issue

gmloose:
Not yet ready to be reviewed and/or merged.


tammojan:
Is the fix you're referring to, to make the interactive shell optional? If so, does that mean that on python 3.8 the interactive shell doesn't work anymore?

gmloose:
No. It basically fixes two issues. So, ideally this should have been two MRs. 
1. PyBDSF doesn't properly build/install for Python >= 3.8. That is the meat of the change.
2. The interactive shell is barely used, but depends on IPython. That dependency was not specified, but should have been. However, IPython drags in a load of stuff, that you probably don't want/need. I therefore made the shell optional.

tammojan:
Is the change to fix 3.8 to replace `scripts` by `entry_points`?

Apart from that, have you tested that the interactive shell still works?

gmloose:
Yes, it still works, if you install the extra of course 😄

setup.py
```diff
@@ -247,27 +247,35 @@ def main():
 
     setup(
         name='bdsf',
-        version='1.9.2',
+        version='1.9.3a1',
```
tammojan, line 250: Please just make this 1.9.3 or 1.9.4.

gmloose:
OK, long story.
This quick-and-dirty fix for Python 3.8 is [just a one-liner](https://github.com/lofar-astron/PyBDSF/issues/168#issuecomment-1021151698). This is the first fix. However, if you try to build `PyBDSF` using a modern PEP-517/518 compliant build tool, it fails. This is because a whole she-bang of sources files are not included in the source tar-ball. This is the second fix, which basically comprises of adding a `MANIFEST.in` file. (Note that there's some duplication now in `setup.py` and `MANIFEST.in`; they now both contain a list of source files.) In order to make the interactive shell an extra, I had to convert the script to an entry point; so that's the third change. And then I updated the documentation, which is the fourth change.

setup.py
gmloose: That's OK. I didn't want to do that until the code was at least reviewed.
You could even argue it should be `1.10` or `1.10.0`, because in the mean time we've already dropped support for Python 2.7, which is still supported by `1.9.2`.

tammojan:
I tested, and the interactive shell is installed also when I just do `pip install --user .`

setup.py
tammojan: The version won't be actually increased until the pull request is merged. `1.10.0` is also fine with me, indeed dropping python 2.7 support is a major thing.

gmloose:
Yep, that's a know "feature" of the default back-end that Python uses to build/install the code. I don't have a solution for that, other than maybe try if (for example) [poetry](https://python-poetry.org/) doesn't do this.

setup.py
gmloose: Do we keep a ChangeLog somewhere? And if so, who's responsible for updating it?

tammojan:
Thanks for the clarification. Great that also the docs are updated.

setup.py
darafferty: Yep, the changelog is at https://github.com/lofar-astron/PyBDSF/blob/master/bdsf/_version.py. I can update it. I'd also like to merge the other open PR (#165) before releasing a new version.

tammojan: Sorry, yes we do keep a changelog. In fact, two:
* bdsf/_version.py
* doc/source/whats_new.rst
We typically update this when a new release is actually released, but you could add changes earlier on.

gmloose:
Ouch, so there are two places where the version number is set: in `setup.py`, and in `bdsf/_version.py`. In Rapthor `setup.py` fetches the version number from the `_version.py` file. Maybe that's something to consider here as well.

tammojan:
I can't find the approve button, but I approve of these changes. Feel free to merge.

setup.py
gmloose: Ouch, so there are two places where the version number is set: in setup.py, and in bdsf/_version.py. In Rapthor setup.py fetches the version number from the _version.py file. Maybe that's something to consider here as well.

gmloose: OK, then it needs a bit more work 😄

PR #173, opened 2022-02-09
Prepare for release of v1.10.0

darafferty:


.travis.yml
```diff
@@ -1,18 +1,17 @@
...
-  - python: 3.5
+  - python: 3.8
 
 before_install:
   - sudo apt-get -qq update
   - sudo apt-get install -y libboost-python-dev gfortran
   - pip install setuptools numpy scipy astropy
 
-install: "python setup.py install"
+install: "pip install ."
```
gmloose, line 15: It may be necessary to update `pip` before using it. Not sure how this should be done in Travis. The command would (obviously) be something like `python3 -m pip install -U pip`

bdsf/_version.py
```diff
@@ -12,6 +12,23 @@ def changelog():
     """
     PyBDSF Changelog.
     -----------------------------------------------------------------------
+    2022/02/09 - Version 1.10.0
+
+    2022/02/09 - Update some functions as required by scipy versions >= 1.8.0
+        (PR #172)
+
+    2022/02/09 - Fix build issues with Python 3.8, end support for Python 2,
```
gmloose, line 20: Also end of support for Python < 3.6. Added support for Python 3.8 and 3.9.

doc/source/whats_new.rst
```diff
@@ -4,10 +4,22 @@
 What's New
 **********
 
+Version 1.10.0 (2022/02/09):
+    * Update some functions as required by scipy version >= 1.8.0 (PR #172)
+
+    * Fix build issues with Python 3.8, end support for Python 2, and make installation
```
gmloose, line 10: Same remark as above.

gmloose (commented):
See my remarks. For the rest, I think you covered the main merge requests, so it's probably complete.

PR #180, opened 2022-10-05
Use len() instead of np.alen().

r-xue:
np.alen() has been removed from NumPy v1.23.

see https://numpy.org/doc/stable/release/1.23.0-notes.html?highlight=alen#expired-deprecations


tammojan:
Thanks!

PR #181, opened 2022-10-05
Add logfilename option

tammojan:
This should make it possible to set a log filename, which is useful if you don't have write access in the directory of the images.


bdsf/opts.py
```diff
@@ -73,6 +73,7 @@ class Opts(object):
     filename = String(doc = "Input image file name\n"\
                                  "The input image can be a FITS or CASA 2-, "\
                                  "3-, or 4-D cube.")
+    logfilename = String(doc = "Log filename")
```
darafferty, line 76: I think it would make more sense to put this option in the "outputs options" section (starting around line 858 in `opts.py`). We should also match the way the other optional string options are defined -- so something like:
```
    logfilename = Option(None, String(),
                             doc = "Filename of the log file. None => get "\
                                 "from filename",
                             group = "output_opts")
```
However, this also raises the issue that if one activates many of the output options (such as `savefits_rmsim` or `output_all`), PyBDSF will still try to write to the input image directory. So perhaps we should make a more general solution, allowing the user to specify the output directory where the log file and other outputs would go? This could mirror the `indir` option that's already there.

tammojan: Ah great idea. I'll have another look.

tammojan: Ah, wait. What would the output directory be, with the current implementation on master, if we specify `indir`? Would it maybe go to the current working directory? That would give a good workaround, without modifying any code

darafferty: Ah, good thought -- unfortunately, as far as I can tell, it is not used that way (see `readimage.py`, around line 45): `indir` seems only to be used for the reading of the image data and header (I can't remember who added this option or why, probably me :) ).

bdsf/functions.py
```diff
@@ -2268,3 +2268,41 @@ def centered(arr, newshape):
...
+        fname = opts['filename']
+    parentname = fname
+
+    # Determine the base directory to use for all output and the log
+    # filename. By default, the log file is place in the same directory
+    # as the input image unless 'outdir' is explicitly defined. We create
+    # the output directory if needed for the log file; otherwise, the
+    # output directory is created in output.py only if needed there
+    if opts['outdir'] is None:
+        basedir = './' + parentname + '_pybdsm/'
```
tammojan, line 2299: pybdsm->pybdsf here?

bdsf/functions.py
```diff
@@ -2268,3 +2268,41 @@ def centered(arr, newshape):
...
+        logfilename = opts['filename'] + '.pybdsf.log'
+    else:
+        basedir = opts['outdir']
+        basename = os.path.basename(opts['filename']) + '.pybdsf.log'
```
tammojan, line 2303: Maybe use `parentname` instead of `opts['filename']` here?

bdsf/functions.py
```diff
@@ -2268,3 +2269,54 @@ def centered(arr, newshape):
     endind = startind + newshape
     myslice = [slice(startind[k], endind[k]) for k in range(len(endind))]
     return arr[tuple(myslice)]
+
+
+def set_up_output_paths(opts):
```
tammojan, line 2274: The name 'base' is a bit confusing here. Could you add in the docstring what `parentname` and `basename` mean? I think a more descriptive name for `basename` would be `output_basedir`? (Intuitively, basename means the basename of the image directory for me.)

darafferty: Yes, good point. I made the suggested changes.

tammojan:
Thanks, looks good to me! I don't have approve powers here, but feel free to merge :)

PR #186, opened 2022-10-31
Add binary wheels for Python 3.10

juliancarrivick:
And bump Boost version to fix removal of deprecated function: https://github.com/boostorg/python/pull/344

I've only bumped the Boost version in the Python 3.10 build, one question would be whether that should be updated for all build wheels.


gmloose:
Ideally yes. But from a practical point of view I'm fine with the current solution.

Issue #192, opened 2023-01-11
numpy 1.24.1 incompatibility - array splitting in `bdsf.multi_proc.parallel_map`


gabicca:
We experience a numpy error when `bdsf.process_image` is called. The error is quite embedded and is coming from `bdsf.multi_proc.parallel_map`.

`parallel_map` calls `numpy.array_split` in line 206. The inputs to this call are two variables, `sequence` and `numcores`. `numcores` is supposedly a number, which is fine. However `sequence` is a list of tuples, the result of the zip function called in `bdsf.rmsimage.Op_rmsimage.rms_mean_map` line 714:

```
cm_cr_list = mp.parallel_map(func.eval_func_tuple,
                    zip(itertools.repeat(self.process_mean_rms_maps),
                    ind_list, itertools.repeat(mask_pad), itertools.repeat(arr_pad),
                    itertools.repeat(kappa)), numcores=ncores)
```

I don't think this is correct. I'm not sure why it worked before, but now numpy will break with the following error:

```
/usr/local/lib/python3.10/site-packages/bdsf/__init__.py:253: in process_image
    img.process(**kwargs)
/usr/local/lib/python3.10/site-packages/bdsf/image.py:135: in process
    success = interface.process(self, **kwargs)
/usr/local/lib/python3.10/site-packages/bdsf/interface.py:62: in process
    _run_op_list(img, op_chain)
/usr/local/lib/python3.10/site-packages/bdsf/__init__.py:152: in _run_op_list
    op(img)
/usr/local/lib/python3.10/site-packages/bdsf/rmsimage.py:263: in __call__
    mean, rms = self.calculate_maps(img, data, mean, rms, mask, map_opts, do_adapt=do_adapt,
/usr/local/lib/python3.10/site-packages/bdsf/rmsimage.py:461: in calculate_maps
    self.map_2d(data, mean, rms, mask, *map_opts, do_adapt=do_adapt,
/usr/local/lib/python3.10/site-packages/bdsf/rmsimage.py:537: in map_2d
    axes, mean_map1, rms_map1 = self.rms_mean_map(arr, mask_small, kappa, box, ncores)
/usr/local/lib/python3.10/site-packages/bdsf/rmsimage.py:714: in rms_mean_map
    cm_cr_list = mp.parallel_map(func.eval_func_tuple,
/usr/local/lib/python3.10/site-packages/bdsf/multi_proc.py:206: in parallel_map
[...]

PR #193, opened 2023-02-08
Fix issues with numpy versions >= 1.24

darafferty:
This PR fixes a few issues that appeared when using recent versions of numpy, including issue #192.

### DIFF
--- a/bdsf/gausfit.py
+++ b/bdsf/gausfit.py
@@ -398,7 +398,7 @@ class Op_gausfit(Op):
             # If all else fails, try to use moment analysis
             if verbose: print('All else has failed')
             inisl = N.where(~isl.mask_active)
-            mask_id = N.zeros(isl.image.shape, dtype=int) - 1
+            mask_id = N.zeros(isl.image.shape, dtype=N.int32) - 1
             mask_id[inisl] = isl.island_id
             try:
                 pixel_beamarea = img.pixel_beamarea()
@@ -645,7 +645,7 @@ class Op_gausfit(Op):
             x, y = N.where(~isl.mask_active); xcen = N.mean(x); ycen = N.mean(y)
             invm = ~isl.mask_active
             #bound = invm - nd.grey_erosion(invm, footprint = N.ones((3,3), int)) # better to use bound for ellipse fitting
-            mom = func.momanalmask_gaus(invm, N.zeros(invm.shape, dtype=int), 0, 1.0, True)
+            mom = func.momanalmask_gaus(invm, N.zeros(invm.shape, dtype=N.int16), 0, 1.0, True)
             g = (maxv, xcen, ycen, mom[3]/fwsig, mom[4]/fwsig, mom[5]-90.)
             gaul.append(g)
             coords.append([xcen, ycen])
--- a/bdsf/gausfit.py
+++ b/bdsf/gausfit.py
@@ -645,7 +645,7 @@ class Op_gausfit(Op):
             x, y = N.where(~isl.mask_active); xcen = N.mean(x); ycen = N.mean(y)
             invm = ~isl.mask_active
             #bound = invm - nd.grey_erosion(invm, footprint = N.ones((3,3), int)) # better to use bound for ellipse fitting
-            mom = func.momanalmask_gaus(invm, N.zeros(invm.shape, dtype=int16), 0, 1.0, True)
+            mom = func.momanalmask_gaus(invm, N.zeros(invm.shape, dtype=int), 0, 1.0, True)
             g = (maxv, xcen, ycen, mom[3]/fwsig, mom[4]/fwsig, mom[5]-90.)
             gaul.append(g)
             coords.append([xcen, ycen])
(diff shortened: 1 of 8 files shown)


bdsf/gausfit.py
```diff
@@ -398,15 +398,15 @@ def fit_island(self, isl, opts, img, ngmax=None, ffimg=None, ini_gausfit=None):
             # If all else fails, try to use moment analysis
             if verbose: print('All else has failed')
             inisl = N.where(~isl.mask_active)
-            mask_id = N.zeros(isl.image.shape, dtype=N.int32) - 1
+            mask_id = N.zeros(isl.image.shape, dtype=int) - 1
```
gmloose, line 401: Here you double the size of `mask_id` , because `dtype=int` is an alias for `dtype=int64`.

bdsf/gausfit.py
```diff
@@ -645,7 +645,7 @@ def inigaus_nobeam(self, isl, thr, beam, img):
             x, y = N.where(~isl.mask_active); xcen = N.mean(x); ycen = N.mean(y)
             invm = ~isl.mask_active
             #bound = invm - nd.grey_erosion(invm, footprint = N.ones((3,3), int)) # better to use bound for ellipse fitting
-            mom = func.momanalmask_gaus(invm, N.zeros(invm.shape, dtype=N.int16), 0, 1.0, True)
+            mom = func.momanalmask_gaus(invm, N.zeros(invm.shape, dtype=int), 0, 1.0, True)
```
gmloose, line 648: Here you quadruple the size of the array of zeros, because `dtype=int` is an alias for `dtype=int64`.

bdsf/islands.py
```diff
@@ -182,7 +182,7 @@ def ndimage_alg(self, img, opts):
         img.island_labels = labels
 
         ### apply cuts on island size and peak value
-        pyrank = N.zeros(image.shape, dtype=N.int32)
+        pyrank = N.zeros(image.shape, dtype=int)
```
gmloose, line 185: Here you double the size of `pyrank`.

bdsf/multi_proc.py
```diff
@@ -203,6 +203,7 @@ def parallel_map(function, sequence, numcores=None, bar=None, weights=None):
     if weights is None or numcores == size:
         # No grouping specified (or there are as many cores as
         # processes), so divide into equal chunks
+        sequence = numpy.array(sequence, dtype=object)
```
gmloose, line 206: Your fix only applies to the `if` part, not the `else` part.
I propose to make the fix in line 161 instead by changing that to:
```
sequence = numpy.array(list(sequence), dtype=object)
```

bdsf/psf_vary.py
```diff
@@ -94,7 +94,7 @@ def __call__(self, img):
             ### now put all relevant gaussian parameters into a list
             ngaus = img.ngaus
             nsrc = img.nsrc
-            num = N.zeros(nsrc, dtype=N.int32)
+            num = N.zeros(nsrc, dtype=int)
```
gmloose, line 97: Here `num` doubles in size.

bdsf/wavelet_atrous.py
```diff
@@ -294,7 +294,7 @@ def __call__(self, img):
               if stop_wav == True:
                   break
 
-          pyrank = N.zeros(img.pyrank.shape, dtype=N.int32)
+          pyrank = N.zeros(img.pyrank.shape, dtype=int)
```
gmloose, line 297: Here `pyrank` doubles in size.

bdsf/wavelet_atrous.py
```diff
@@ -658,7 +658,7 @@ def renumber_islands(img):
 
     Also renumbers the pyrank image.
     """
-    pyrank = N.zeros(img.pyrank.shape, dtype=N.int32)
+    pyrank = N.zeros(img.pyrank.shape, dtype=int)
```
gmloose, line 661: See line 297

bdsf/islands.py
```diff
@@ -114,7 +114,7 @@ def __call__(self, img):
 
             ch0_map = img.ch0_arr
             ch0_shape = ch0_map.shape
-            pyrank = N.zeros(ch0_shape, dtype=N.int32)
+            pyrank = N.zeros(ch0_shape, dtype=int)
```
gmloose, line 117: Here `pyrank` doubles in size.

gmloose:
BTW: the addition of `dtype=object` indeed fixes the issue.

bdsf/gausfit.py
darafferty: Yeah, I realized that, but it shouldn't matter in practice (the mask is not a large array).

bdsf/gausfit.py
darafferty: Indeed, but the array is small in practice (as above).

bdsf/islands.py
darafferty: Here the array is larger, so it might be more important (though I don't think it will make any practical difference here either). I will change all the occurrences back to their original values just to be on the safe side.

bdsf/multi_proc.py
darafferty: Good catch. I will use your suggested fix.

gmloose:
LGTM.

PR #194, opened 2023-02-10
Prepare 1.10.2 release

gmloose:


bdsf/_version.py
```diff
@@ -4,14 +4,30 @@
...
 
 
 # Changelog
 def changelog():
     """
     PyBDSF Changelog.
     -----------------------------------------------------------------------
+    2023/02/10 - Version 1.10.2
+
+    2023/02/10 - Fix issues with numpy versions >= 1.24 (#193)
```
darafferty, line 17: I just realized it was issue #192, not #193 as I wrote in the PR. Can you change it here and in `whats_new.rst`?

doc/source/whats_new.rst
```diff
@@ -4,6 +4,21 @@
 What's New
 **********
 
+Version 1.10.2 (2023/02/10):
+    * Fix issues with numpy versions >= 1.24 (#193)
```
darafferty, line 8: As in the changelog, it should be #192 here.

bdsf/_version.py
darafferty: Actually, #193 was a reference to the PR, so it is fine here.

darafferty:
Looks good

PR #196, opened 2023-03-27
Replace Travis CI with GitHub actions

gmloose:
Replace Travis CI, which is currently broken and outdated, with GitHub actions.


darafferty:
Looks good! Great to have CI working again.

Issue #195, opened 2023-03-22
Matplotlib==3.7.0 Atttibute Error


mohini2110:
Using interactive shell with matplotlib latest version(3.7.0) I got the following error

`AttributeError: 'FigureCanvasQTAgg' object has no attribute 'set_window_title'` 

However, downgrading to matplotlib==3.4.0 worked with a Deprecation Warning:

`The set_window_title function was deprecated in Matplotlib 3.4 and will be removed two minor releases later. Use manager.set_window_title or GUI-specific methods instead.
  fig.canvas.set_window_title('PyBDSM Fit Results for '+ img.filename)
/path_to_python3.8/site-packages/bdsf/plotresults.py:252: MatplotlibDeprecationWarning: You are modifying the state of a globally registered colormap. This has been deprecated since 3.3 and in 3.6, you will not be able to modify a registered colormap in-place. To remove this warning, you can make a copy of the colormap first. cmap = mpl.cm.get_cmap("gray").copy()
  gray_palette.set_bad('k')`

gmloose:
Thanks for the report. We will look into this. Deprecation warnings often stay below the radar, unfortunately.

PR #199, opened 2023-03-31
Fix #198

### LINKED ISSUE #198: canvas.set_window_title deprecated
Matplotlib 3.4 deprecated `canvas.set_window_title` causing `show_fit()` to no longer work with recent matplotlib versions.

tikk3r:
Fixes #198 by first trying the old method and if that fails trying the updated call to `canvas.manager.set_window_title` instead of `canvas.set_window_title`.

### DIFF
--- a/bdsf/plotresults.py
+++ b/bdsf/plotresults.py
@@ -247,7 +247,12 @@ def plotresults(img, ch0_image=True, rms_image=True, mean_image=True,
         numx = 1
     numy = int(N.ceil(float(len(images))/float(numx)))
     fig = pl.figure(figsize=(max(15, 10.0*float(numy)/float(numx)), 10.0))
-    fig.canvas.set_window_title('PyBDSM Fit Results for '+ img.filename)
+    try:
+        # This way has been deprecated with Matplotlib 3.4.
+        fig.canvas.set_window_title('PyBDSM Fit Results for '+ img.filename)
+    except AttributeError:
+        # We are on a newer Matplotlib where the above is deprecated.
+        fig.canvas.manager.set_window_title('PyBDSM Fit Results for '+ img.filename)
     gray_palette = cm.gray
     gray_palette.set_bad('k')
 


tikk3r:
Should also fix #195 since #198 seems to be a duplicate of that.

bdsf/plotresults.py
```diff
@@ -247,7 +247,12 @@ def plotresults(img, ch0_image=True, rms_image=True, mean_image=True,
         numx = 1
     numy = int(N.ceil(float(len(images))/float(numx)))
     fig = pl.figure(figsize=(max(15, 10.0*float(numy)/float(numx)), 10.0))
-    fig.canvas.set_window_title('PyBDSM Fit Results for '+ img.filename)
+    try:
+        # This way has been deprecated with Matplotlib 3.4.
+        fig.canvas.set_window_title('PyBDSM Fit Results for '+ img.filename)
+    except AttributeError:
+        # We are on a newer Matplotlib where the above is deprecated.
+        fig.canvas.manager.set_window_title('PyBDSM Fit Results for '+ img.filename)
```
gmloose, line 255: I would prefer the logic to be reversed. In the sense that the happy path is with latest Matplotlib, and the unhappy path with old versions. Or..., assuming we can easily check the version of Matplotlib, use an `if/else` construct.

tammojan: The new behavior already works in matplotlib 1.2.0, released 2021, and the default in centos 7. I suggest to just use the new way and get rid of the try/except.

tikk3r: Right, if it has been in there that long already I would assume most people are somewhat up to date and that the new way is probably safe. I've changed it to the new way in that case.

gmloose: Fine with me. Please add a minimal version requirement for `matplotlib` in `setup.py`.
Oh, I see that `matplotlib` doesn't even appear at all in the `install_requires`. Not sure why. Probably because it's only needed when using the extra `[ishell]`. I always find `pip install` and `setup.py` a bit awkward when it comes to extras. So, maybe just leave `setup.py` as is.

PR #203, opened 2023-05-02
Use cibuildwheel to build binary wheels

gmloose:
Use `cibuildwheel` to build binary wheels for both Linux and MacOS (Intel).


tammojan (approved):

gmloose:
> This looks great to me! Please make sure to squash the commits when merging.
> 
> I guess the final test of the system is to make a release tag, feel free to do so. For that, we may need to add some more entries in the changelog (of changes that happened on master recently).

I already tested that. You can find a pre-release 1.10.3a2 on PyPI. The ChangeLog should indeed be updated when we make a new release.

PR #205, opened 2023-05-08
Fix build issue with Python 3.11

### LINKED ISSUE #189: Debian build for Python 3.11 fails
This is a forward of [Debian#1024499](https://bugs.debian.org/1024499):

When trying to build the package for Python 3.11, the build fails with the following error when trying to import bdsf:
```
ERROR: bdsf (unittest.loader._FailedTest.bdsf)
----------------------------------------------------------------------
ImportError: Failed to import test module: bdsf
Traceback (most recent call last):
  File "/usr/lib/python3.11/unittest/loader.py", line 440, in _find_test_path
    package = self._get_module_from_name(name)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/unittest/loader.py", line 350, in _get_module_from_name
    __import__(name)
  File "/<<PKGBUILDDIR>>/.pybuild/cpython3_3.11/build/bdsf/__init__.py", line 21, in <module>
    from .collapse import Op_collapse
  File "/<<PKGBUILDDIR>>/.pybuild/cpython3_3.11/build/bdsf/collapse.py", line 11, in <module>
    from . import _cbdsm
SystemError: initialization of _cbdsm raised unreported exception
```
This happens on all our platforms. I have no idea why this "unreported exception" is raised. Full build log for x86_64 is [here](https://buildd.debian.org/status/fetch.php?pkg=pybdsf&arch=amd64&ver=1.10.1-1%2Bb3&stamp=1668731964&raw=0).

Any idea what this could be?

gmloose:
The Python 3.11 build failed, due to some obscure error generated by `f2py`. 

Further investigation showed that the problem is caused by a bug in `f2py.crackfortran`: see https://github.com/numpy/numpy/issues/23533 (BUG: f2py.crackfortran fails on comment after endif statement). The quick work-around was to fix the affected Fortran files.

After that fix, it turned out that the self test failed. This could be tracked down to a problem with Boost and Python 3.11: see https://github.com/boostorg/python/issues/400 (Python 3.11 gives SystemError: type Boost.Python.enum has the Py_TPFLAGS_HAVE_GC flag but has no traverse function). Bumping the Boost version to 1.81 resolved this issue.

Fixes #189

### DIFF
--- a/manylinux2014/build.sh
+++ b/manylinux2014/build.sh
@@ -8,7 +8,7 @@ SCRIPT_DIR=$(cd $(dirname $0) && pwd)
 ROOT_DIR=$(git rev-parse --show-toplevel)
 
 py_major=3
-for py_minor in $(seq 6 10); do
+for py_minor in $(seq 6 11); do
     echo -e "\n\n******** Building wheel for python ${py_major}.${py_minor} ********\n"
     py_version=${py_major}${py_minor}
     [ ${py_minor} -le 7 ] && py_unicode="m" || py_unicode=
--- a/manylinux2014/py_wheel.docker
+++ b/manylinux2014/py_wheel.docker
@@ -18,9 +18,9 @@ RUN /opt/python/${TARGET}/bin/pip install oldest-supported-numpy "setuptools<65.
 
 # download and build boost
 WORKDIR /build
-RUN curl https://ufpr.dl.sourceforge.net/project/boost/boost/1.76.0/boost_1_76_0.tar.bz2 --output boost.tar.bz2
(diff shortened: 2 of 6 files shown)


darafferty:
Great find! I've no idea why this works either.

darafferty (approved):

gmloose:
> Great find! I've no idea why this works either.

Well, apparently it doesn't :(
The test now fails with the following error:
```
AttributeError: module 'bdsf._pytesselate' has no attribute 'pytess_simple'
```
So, I have to dig into this a bit deeper I'm afraid.

darafferty (approved):
Looks good!

Issue #200, opened 2023-04-20
Wavelet mode failing if output_all=True


halecatherine:
When I am running PyBDSF on a test image with imaging commands which include the atrous fitting mode, it is running fine, however if I include the output_all = True option, then it processes the image but is failing on the first wavelet mode as it is unable to find an input directory and has a filename which does not appear to agree with files saved. 

```>>> img = bdsf.process_image('J105007+304037.fits', detection_image='J105007+304037.fits', thresh_isl=4.0, thresh_pix=5.0, rms_box=(150,15), rms_map=True, mean_map='zero', ini_method='intensity', adaptive_rms_box=True, adaptive_thresh=150, rms_box_bright=(60,15), group_by_isl=False, group_tol=10.0, output_opts=True, output_all=True, atrous_do=True, atrous_jmax=4, flagging_opts=True, flag_maxsize_fwhm=0.5, advanced_opts=True, blank_limit=None, frequency=1400000000)
--> Opened 'J105007+304037.fits'
Image size .............................. : (667, 667) pixels
Number of channels ...................... : 1
Number of Stokes parameters ............. : 1
Beam shape (major, minor, pos angle) .... : (1.50000e-03, 1.50000e-03, 0.0) degrees
Frequency of image ...................... : 1400.000 MHz
Number of blank pixels .................. : 0 (0.0%)
Flux from sum of (non-blank) pixels ..... : 0.325 Jy
--> Calculating background rms and mean images
--> Using adaptive scaling of rms_box
Using user-specified rms_box ............ : (60, 15) pixels (small scale)
Using user-specified rms_box ............ : (150, 15) pixels (large scale)
Number of sources using small scale ..... : 1
--> Using 2D map for background rms
--> Using constant background mean
Min/max values of background rms map .... : (1.33e-04, 1.91e-04) Jy/beam
Value of background mean ................ : 0.0 Jy/beam
--> Expected 5-sigma-clipped false detection rate < fdr_ratio
--> Using sigma-clipping ('hard') thresholding
Minimum number of pixels per island ..... : 6

--> Determining islands from detection image
--> Opened 'J105007+304037.fits'
Image size .............................. : (667, 667) pixels
Number of channels ...................... : 1
Number of Stokes parameters ............. : 1
Beam shape (major, minor, pos angle) .... : (1.50000e-03, 1.50000e-03, 0.0) degrees
Frequency of image ...................... : 1400.000 MHz
[...]

PR #209, opened 2023-06-21
Add options for use with a detection image

darafferty:
This PR adds an option to allow user-supplied rms and mean maps to be used with a detection image. Options were also added to allow the rms and mean maps generated when a detection image is used to be saved. Also fixes issue #200.


bdsf/islands.py
```diff
@@ -100,6 +105,26 @@ def __call__(self, img):
...
             img.minpix_isl = det_img.minpix_isl
+
+            if opts.output_all:
+                write_islands(img)
+            if opts.savefits_rankim or opts.output_all:
+                func.write_image_to_file(img.use_io, img.imagename + '_pyrank.fits', img.pyrank, img)
+            if opts.savefits_det_rmsim or opts.output_all:
+                resdir = img.basedir + '/background/'
+                if not os.path.exists(resdir):
+                    os.makedirs(resdir)
```
gmloose, line 116: Not sure if this execution path can be taken multiple times, but `os.makedirs` will raise an exception if the directory already exists. To avoid this, you need to add the option `exist_ok=True`.

bdsf/islands.py
```diff
@@ -100,6 +105,26 @@ def __call__(self, img):
...
+                func.write_image_to_file(img.use_io, img.imagename + '.detection_rmsd_I.fits',
+                                         img.detection_rms_arr, img, resdir)
+                mylog.info('%s %s' % ('Writing ', resdir+img.imagename+'.detection_rmsd_I.fits'))
+            if opts.savefits_det_meanim or opts.output_all:
+                resdir = img.basedir + '/background/'
+                if not os.path.exists(resdir):
+                    os.makedirs(resdir)
```
gmloose, line 123: And here you could use the `exist_ok=True`, so you don't have to check for existence first.

gmloose (approved):
LGTM. I see a lot of options, but these appear to be all well documented, so I guess that's OK.

gmloose:
I would recommend running `pylint` on your sources and at least fix all warnings.

Issue #201, opened 2023-04-26
Spectral index fitting to empty-channeled cube with frequency-dependent synthesized beam


rstofi:
I am trying to run `pybdsf` on the output from WSclean MFS images. I made a cube from the primary beam (PB) corrected sub-band images and have the non-PB-corrected MFS image. I use the latter for source-finding, and the cube for measuring sources including the spectral index calculation. However, I ran into several issues with this approach, in particular, I have a fully-flagged channel in my MFS cube and am not sure if I am handling the frequency-dependent synthesized beams correctly.

First, when I specify the `collapse_av` and `beam_spectrum` parameters to deal with the frequency-dependent beams and the empty channel, `pybdsf` still seems to use the beam from the header of the cube (I needed to add a beam to the header as `pybdsf` cannot finds the frequency-dependent beams from a second HDU `BINTABLE` from a fits file). Providing the beams literally seems to have no effect. I am not sure then how the sub-band beams are used then, and I could not have figured it out yet...

Related to this is the fact that WSclean provides the weights to collapse the sub-band images into the MFS image (the `ch0_image`), but in `pybdsf` the weights are based on the measured rms of the sub-bands. I found no potion to provide my own weights for collapsing the cube, but I found that there should be an `image` mode for the `collapse_mode` parameter when multichannel options are enabled. However, when I set this and provide the PB-corrected MFS image to the `collapse_file`  parameter, I got the following error:

```
RuntimeError: Parameter "collapse_wt" is not defined properly. 
 ("Failed to set property collapse_wt of class Opts to a value of 'file'; expected a value of 'unity' or 'rms'.", "'file'", "a value of 'unity' or 'rms'", 'collapse_wt', 'Opts')
```

This function seems to be in the code, but not in the documentation. I need help to make this work, so I have no problems with the different collapsing methods...

I want to measure the spectral indices of the sources since I have a wide-band image, and so I don't want to use the MFS images for measurement but the cube. However, when I enable the `spectralindex_do` parameter (with the `collapse_av` set to avoid the empty channels in my cube), I got the following error:

```
RuntimeError: A region with an unphysical rms value has been found. Please check the input image.
```

So it seems that the empty channel is not ignored when spectral index calculations are enabled. Is this a known/expected behaviour, or do I have an issue with my parameter setup? I attached a log file when I ran into this issue.

I am not an expert in using `pybdsf`, but I've been told that what I want to do is feasible with it. Which is:
[...]

PR #210, opened 2023-06-21
Fix various issue with the spectral index module

darafferty:
This PR fixes a number of issues in the spectral index module, most of which were related to the the lack of channel-specific masking. An option was also added to allow the user to specify channels to flag before spectral indices are calculated. Fixes issue #201.


bdsf/spectralindex.py
```diff
@@ -60,7 +60,7 @@ def __call__(self, img):
                 unav_freqs = freqin[good_chans]
                 nmax_to_avg = img.opts.specind_maxchan
                 nchan = unav_image.shape[0]
-                mylog.info('After initial flagging of channels by rms, %i good channels remain' % (nchan,))
+                mylogger.userinfo(mylog, 'Number of channels remaining after initial flagging', str(nchan))
```
gmloose, line 63: Curious as to why you changed this from `mylog.info` to `mylogger.userinfo`.

gmloose (approved):
You might want to run your sources through `pylint` and at least fix all warnings.

bdsf/spectralindex.py
darafferty: That was so that the message is printed to the terminal and not only to the log file.

darafferty:
I fixed the `pylint` warnings -- thanks for suggesting that.

PR #211, opened 2023-07-12
Fix issues with error estimation

darafferty:
This PR fixes a number of issues with error estimation, including an incorrect treatment of the errors when correcting for projection effects and when the `fix_to_beam` option is enabled.

### DIFF
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -793,7 +793,7 @@ def get_errors(img, p, stdav, bm_pix=None):
         dumrr2 = 1.0 + bm_pix[0] * bm_pix[1] / (size[1] * size[1])
         dumrr3 = dumr * pp[0] / stdav
         d1 = sqrt(8.0 * log(2.0))
-        d2 = (size[0] * size[0] - size[1] * size[1]) / (size[0] * size[0])
+        d2 = (size[0] * size[0] - size[1] * size[1]) / (size[0] * size[1])
         try:
             e_peak = pp[0] * sq2 / (dumrr3 * pow(dumrr1, 0.75) * pow(dumrr2, 0.75))
             e_maj = size[0] * sq2 / (dumrr3 * pow(dumrr1, 1.25) * pow(dumrr2, 0.25))
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -753,7 +753,7 @@ def deconv2(gaus_bm, gaus_c):
     return (bmaj, bmin, bpa), ifail
 
 
-def get_errors(img, p, stdav, bm_pix=None):
+def get_errors(img, p, stdav, bm_pix=None, fixed_to_beam=False):
     """ Returns errors from Condon 1997
 
     Returned list includes errors on:
@@ -812,7 +812,17 @@ def get_errors(img, p, stdav, bm_pix=None):
             e_min = 0.0
             e_pa = 0.0
             e_tot = 0.0
-        if abs(e_pa) > 180.0: e_pa=180.0  # dont know why i did this
+        if abs(e_pa) > 180.0:
+            e_pa = 180.0
+        if fixed_to_beam:
+            # When the size was fixed to that of the beam during the fit, set
+            # uncertainties on the size to zero and reduce the error in the fluxes
+            # by sqrt(2) (see Eq. 25 of Condon 1997)
+            e_maj = 0.0
+            e_min = 0.0
+            e_pa = 0.0
+            e_peak /= sq2
+            e_tot /= sq2
         errors = errors + [e_peak, e_x0, e_y0, e_maj, e_min, e_pa, e_tot]
 
     return errors
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -754,7 +754,17 @@ def deconv2(gaus_bm, gaus_c):
 
 
 def get_errors(img, p, stdav, bm_pix=None, fixed_to_beam=False):
-    """ Returns errors from Condon 1997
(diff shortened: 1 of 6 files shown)


bdsf/functions.py
```diff
@@ -753,7 +753,7 @@ def deconv2(gaus_bm, gaus_c):
     return (bmaj, bmin, bpa), ifail
 
 
-def get_errors(img, p, stdav, bm_pix=None):
+def get_errors(img, p, stdav, bm_pix=None, fixed_to_beam=False):
```
gmloose, line 756: There's no documentation for the new `fixed_to_beam` argument. Also, you may want to make clear what kind of errors you're referring to exactly. I guess numerical errors, but that's not clear from the function name and the documentation. Also Condon 1997 may need a reference for the uninformed like me :).

bdsf/functions.py
```diff
@@ -793,7 +793,7 @@ def get_errors(img, p, stdav, bm_pix=None):
         dumrr2 = 1.0 + bm_pix[0] * bm_pix[1] / (size[1] * size[1])
         dumrr3 = dumr * pp[0] / stdav
         d1 = sqrt(8.0 * log(2.0))
-        d2 = (size[0] * size[0] - size[1] * size[1]) / (size[0] * size[0])
+        d2 = (size[0] * size[0] - size[1] * size[1]) / (size[0] * size[1])
```
gmloose, line 796: Oh, that's an interesting catch.
BTW, not related to this MR, this piece of code could benefit from some inline comments explaining what's being done, and/or a bit more descriptive variable names.

bdsf/gausfit.py
```diff
@@ -1022,10 +1022,10 @@ def __init__(self, img, gaussian, isl_idx, g_idx, flg=0):
                 func.approx_equal(size[1], img.pixel_beam()[1]) and \
                 func.approx_equal(size[2], img.pixel_beam()[2]+90.0) or \
                 img.opts.fix_to_beam:
-            # Check whether fitted Gaussian is just the distorted pixel beam
-            # given as an initial guess or if size was fixed to the beam. If so,
-            # reset the size to the undistorted beam.
-            # Note: these are sigma sizes, not FWHM sizes.
+            # Check whether fitted Gaussian is just the distorted pixel beam given as an
+            # initial guess (i.e., [bm_maj*1.1, bm_min, bm_pa+90]) or if size was fixed to
```
gmloose, line 1026: `i.e.` or `e.g.`? In other words: "this is", or "for example"?

bdsf/readimage.py
```diff
@@ -233,32 +233,38 @@ def gaus2pix(x, location=None, use_wcs=True):
...
-        def pix2gaus(x, location=None, use_wcs=True):
+        def pix2gaus(x, location=None, use_wcs=True, is_error=False):
             """ Converts Gaussian parameters in pixels to deg.
 
             x - (maj [pix], min [pix], pa [deg])
             location - specifies the location in pixels (x, y) for which
                 transform is desired
             Input beam angle should be degrees CCW from the +y axis of the image.
             The output beam angle is degrees CCW from North.
+            Set is_error = True when x gives the errors on the parameters instead of
```
gmloose, line 244: Does "gives" mean "contains" here?

bdsf/readimage.py
```diff
@@ -233,32 +233,38 @@ def gaus2pix(x, location=None, use_wcs=True):
...
+                bpa = th
+                if not is_error:
+                    # Adjust the PA by the rotation delta and fix cases where
+                    # major and minor axes are swapped
+                    brot = self.get_rot(img, location) # rotation delta CCW (in degrees) between N and +y axis of image
+                    bpa = th - brot
+                    if bmaj < bmin:
+                        bmaj, bmin = bmin, bmaj
+                        bpa += 90.0
+                    bpa = divmod(bpa, 180)[1] ### bpa lies between 0 and 180
```
gmloose, line 264: Is there no `else` clause, or did you forget to add it?

bdsf/readimage.py
```diff
@@ -329,21 +335,24 @@ def beam2pix(x):
...
             bpa = th
-            if bmaj < bmin:
-                bmaj, bmin = bmin, bmaj
-                bpa += 90.0
-            bpa = divmod(bpa, 180)[1] ### bpa lies between 0 and 180
+            if not is_error:
+                if bmaj < bmin:
+                    bmaj, bmin = bmin, bmaj
+                    bpa += 90.0
+                bpa = divmod(bpa, 180)[1] ### bpa lies between 0 and 180
```
gmloose, line 355: This code snippet also appeared in `bdsf/readimage.py`. Maybe turn it into a separate utility function?

bdsf/functions.py
darafferty: Good points -- I updated the docstring with more info.

bdsf/functions.py
darafferty: Very true. I always have to search for a good while to figure out exactly where these come from. I added comments for each variable as its origin.

bdsf/readimage.py
darafferty: Yes. I changed it to "contains" to be clear.

bdsf/readimage.py
darafferty: No, no other processing is needed in that case.

bdsf/readimage.py
darafferty: Good point -- I see it three times in fact. I made a function for it (in `functions.py`).

gmloose:
LGTM

Issue #212, opened 2023-09-07
savefits_residim does not work


PaddyLeahy:
I set the savefits_residim option to True in process_image, but no such fits image is output. In the log file, this is not listed as a non-default input parameter (unlike savefits_rmsim, which is, and does produce an output image.  I realize that I can use export_image to do this, but having the residual image generated automatically, in the appropriate subdirectory like the rms image, would be convenient!

darafferty:
Thanks for reporting the problem. Indeed, the option was completely ignored. I've made a PR (https://github.com/lofar-astron/PyBDSF/pull/213) to fix it. The PR also adds an option (`savefits_modelim`) for saving the Gaussian model image, since that was an easy related change.

PaddyLeahy:
Hi David,
      Always happy to report bugs! I've recently returned to my pybdsf project and I have a query: where in the code is the total flux evaluated for gaussians? I've looked on github without success, but I assume this is not done deep in the gaussian fitting process as the total flux is not an obvious fit parameter. I ask because I have noticed that in some but not all circumstances the total flux is set equal to the peak flux if the deconvolved size is zero, not just when the fitted size is exactly equal to the model beam, but (usually) when it is smaller. This happened in a run with ini_gausfit = 'nobeam', if that's relevant. More usually, total flux < peak flux for sources with fitted size < beam. It would be useful to know exactly what triggers the switch.

Regards,
            Paddy
________________________________

darafferty:
The total flux is calculated here: https://github.com/lofar-astron/PyBDSF/blob/master/bdsf/gausfit.py#L1043. There's a check just above (line 1021) that might be relevant for the issue you're seeing. It sets the size to the beam size in certain cases. But I don't see any obvious relation between what's done here and the value of `ini_gausfit`. However, the logic surrounding all that is quite convoluted so it may be related somehow. The deconvolution is done just after the calculation of the total flux in line 1053.

PaddyLeahy:
Thanks for the quick reply!
________________________________

PR #213, opened 2023-09-11
Fix saving of residual image and add option to save model image

darafferty:
This PR fixes issue #212 and adds a new option to `process_image` (called `savefits_modelim`) to save the Gaussian model image as well.


gmloose:
LGTM

PR #216, opened 2024-03-07
Replace distutils for Python 3.12

gmloose:
Python 3.12 no longer ships with `distutils`. Replaced the use of `distutils.version` with `packaging.version`.


darafferty (approved):
Looks good!

PR #222, opened 2024-04-19
Rap 612 remove pyfits version checks

gmloose:
Remove checks and alternative code paths for old `pyfits` and `pywcs`. These packages are part of `astropy` for quite some time already.


bdsf/functions.py
```diff
@@ -1120,80 +1125,36 @@ def read_image_from_file(filename, img, indir, quiet=False):
...
-        img._reason = 'File does not exist'
+        img._reason = f'File {image_file} does not exist'
         return None
 
     # If img.use_io is set, then use appropriate io module
-    if img.use_io != '':
+    if img.use_io:
+        # Sanity check: only 'fits' and 'rap' are supported image I/O types
+        if img.use_io not in ('fits', 'rap'):
+            raise ValueError("Invalid image I/O type '{img.use_io}'. "
```
darafferty, line 1135: I think we need an f-string here?

darafferty (commented):
Wow, there was a lot of stuff in there! Nice to have it gone.

bdsf/functions.py
gmloose: Good catch. Otherwise the sanity check would become a source of error.

darafferty (approved):
Looks good!

PR #224, opened 2024-05-06
Rap 614 use setuptools scm

gmloose:
Use `setuptools_scm` to do automatic versioning.

Since `setuptools_scm` recommends to use `pyproject.toml` instead of `setup.py`, most setting were move from `setup.py` to `pyproject.toml`. However, `scikit-build` requires a few settings to be set in `setup.py`. So these were retained.


gmloose:
The issue is not with `setuptools_scm`, but with` scikit-build`. Once `scikit-build-core` provides proper support for `f2py`, we can probably switch. The advantage of `scikit-build-core` is that is integrates `setuptools_scm` very nicely.

PR #226, opened 2024-05-17
Cleanup

AlexKurek:


gmloose:
@AlexKurek, you may want to review your latest commit, because the tests no longer pass after that.

AlexKurek:
> @AlexKurek, you may want to review your latest commit, because the tests no longer pass after that.

@gmloose I think I fixed it in https://github.com/lofar-astron/PyBDSF/pull/226/commits/4937761a3b88f40f712926dff38d15786dd93447.

PR #228, opened 2024-06-11
Output island rms and mean values for wavelet Gaussians

darafferty:
This PR adds two columns to the output Gaussian catalogs: `Wave_Isl_rms` and `Wave_Isl_mean`. These columns give the values of the background rms and mean of the wavelet image of the island to which a given Gaussian belongs.


gmloose (commented):
Looks good to me.

Issue #227, opened 2024-06-09
Vectorization of Fortran loops


AlexKurek:
Just a question / suggestion: this are the results of profiling:

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      571 7110.513   12.453 7110.513   12.453 {built-in method posix.waitpid}
        4 1365.157  341.289 2005.113  501.278 wavelet_atrous.py:667(check_islands_for_overlap)
        1  830.799  830.799  830.799  830.799 psf_vary.py:685(tess_simple)
     3407  381.475    0.112  381.497    0.112 necompiler.py:978(re_evaluate)
     1544  312.586    0.202  312.586    0.202 {method 'partition' of 'numpy.ndarray' objects}
       20  305.964   15.298  305.964   15.298 {built-in method scipy.ndimage._nd_image.geometric_transform}
       44  288.652    6.560  288.652    6.560 {built-in method gc.collect}
        5  270.768   54.154 1012.391  202.478 functions.py:2223(bstat)
      214  168.266    0.786  190.097    0.888 wavelet_atrous.py:649(renumber_islands)
   314391  137.325    0.000  225.320    0.001 _methods.py:196(_var)
  5110219  134.055    0.000  134.055    0.000 {method 'reduce' of 'numpy.ufunc' objects}
        6  117.511   19.585  251.401   41.900 make_residimage.py:25(__call__)
    27523  106.675    0.004  106.675    0.004 {method 'flatten' of 'numpy.ndarray' objects}
10056453/7133976   97.635    0.000  817.598    0.000 {built-in method numpy.core._multiarray_umath.implement_array_function}
    31858   55.493    0.002   55.493    0.002 {built-in method posix.read}
```
Have you tried to vectorize loops there:
https://github.com/lofar-astron/PyBDSF/blob/master/src/fortran/pytess_simple.f#L16C1-L29C15
?
ChatGPT seems to be doing this:
```
program vectorized_code
  implicit none
  integer, parameter :: m = 100, n = 100, ngens = 10
  integer :: i, j, k
  real(8), dimension(n, m) :: volrank
  real(8), dimension(ngens) :: xgens, ygens, wts
  real(8), dimension(n, m, ngens) :: dist
  real(8) :: dumr
  integer, dimension(n, m) :: minind

  ! Initialize generators and weights (example values)
  do k = 1, ngens
     xgens(k) = k
     ygens(k) = k
     wts(k) = 1.0d0
  end do

[...]

PR #231, opened 2024-06-20
Add checks of output to tests

darafferty:
This PR adds checking of the outputs to the main test script (see RAP-694; it is also relevant to issue #227). The new tests compare the outputs to a reference set (stored in `tests/reference_outputs`). Currently, it is checked that:

- all expected files are present
- the noise in each output image agrees (within 1%) with that of the corresponding reference image

The contents of text files and catalogs are not currently checked, as the details of the fitting (e.g., Gaussian properties, number of Gaussians, their ordering, etc.) may vary due to differences in the libraries, etc. used between the test run and the run used to generate the reference outputs.

Some cleanup was also done, including the removal of old, unused test files.


test/compare_outputs.py
```diff
@@ -0,0 +1,417 @@
...
+    :param str file2: Second text file
+    :param int verbosity: Verbosity level, higher is more verbose
+    :return bool: True if text files are similar, else False
+    """
+    logger.debug("Comparing '%s' and '%s'", file1, file2)
+    d = difflib.Differ()
+    with open(file1) as f1, open(file2) as f2:
+        diff = list(d.compare(f1.readlines(), f2.readlines()))
+    agree = True
+    for line in diff:
```
gmloose, line 65: I would filter `diff` already now, since you need it both directly below and in the log (line 77). Something like:
```
    # Only keep lines that start with diff markers and are not comment lines
    diff = [
        line for line in d.compare(f1.readlines(), f2.readlines()) 
        if line.startswith(('-','+')) and not line[2:].startswith('#')
    ] 
```

test/compare_outputs.py
```diff
@@ -0,0 +1,417 @@
...
+                        agree = False
+            if check_other:
+                if not text_files_are_similar(
+                    left, right, verbosity=verbosity
+                ):
+                    agree = False
+    return agree
+
+
+class FITSImage(object):
```
gmloose, line 212: Please add a note that this class was copied (almost) verbatim from the Rapthor repository: `rapthor/lib/fitsimage.py`. Maybe add a hyperlink to this file on GitLab as well?

gmloose (commented):
Looks good to me, except for some minor things. See my comments.

gmloose:
One more thing. I see that the loglevel is set to `DEBUG` for the comparison in `test_bdsf_process_image.py`, which creates a lot of noise. Could the level be set to `INFO` or `WARNING` by default?

gmloose:
O, and when I try to run the test it fails with the following error:
```
ERROR: The following files are present in 'reference_outputs/misc' but not in 'tbdsf_process_image.in_fits_pybdsf/misc': ['tbdsf_process_image.in_fits.pybdsf.atrous.w1.pybdsf_pyrank.fits', 'tbdsf_process_image.in_fits.pybdsf.atrous.w2.pybdsf_pyrank.fits', 'tbdsf_process_image.in_fits.pybdsf.atrous.w3.pybdsf_pyrank.fits', 'tbdsf_process_image.in_fits.pybdsf.atrous.w4.pybdsf_pyrank.fits', 'tbdsf_process_image.in_fits.pybdsf_pyrank.fits']
```

gmloose:
> O, and when I try to run the test it fails with the following error:
> 
> ```
> ERROR: The following files are present in 'reference_outputs/misc' but not in 'tbdsf_process_image.in_fits_pybdsf/misc': ['tbdsf_process_image.in_fits.pybdsf.atrous.w1.pybdsf_pyrank.fits', 'tbdsf_process_image.in_fits.pybdsf.atrous.w2.pybdsf_pyrank.fits', 'tbdsf_process_image.in_fits.pybdsf.atrous.w3.pybdsf_pyrank.fits', 'tbdsf_process_image.in_fits.pybdsf.atrous.w4.pybdsf_pyrank.fits', 'tbdsf_process_image.in_fits.pybdsf_pyrank.fits']
> ```
> 
> Which is kind of curious, because these files seem to be present. 🤔

I already see what's the problem. The missing outputs are not in `tbdsf_process_image.in_fits_pybdsf/misc`, but in `.`.

darafferty:
That's strange -- the files are in the right place (in `misc/`) for me (and in the CI pipeline as well I guess, since it passes the check). The output directory was changed to `misc/` in commit dfe265a1844a79c0aebb5a84c7db3fe46dd41780.

darafferty:
Ah, I found the problem I think: it could write the `pyrank` images to the wrong directory in some cases. Should be fixed now.

bdsf/islands.py
```diff
@@ -142,7 +142,10 @@ def __call__(self, img):
             if opts.output_all:
                 write_islands(img)
             if opts.savefits_rankim or opts.output_all:
-                func.write_image_to_file(img.use_io, img.imagename + '_pyrank.fits', pyrank, img)
+                resdir = img.basedir + '/misc/'
```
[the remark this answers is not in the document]
darafferty: Good idea. Done.

bdsf/islands.py
```diff
@@ -142,7 +142,10 @@ def __call__(self, img):
...
+                os.makedirs(resdir, exist_ok=True)
+                func.write_image_to_file(img.use_io, img.imagename + '_pyrank.fits', pyrank, img, resdir)
+                mylog.info('%s %s' % ('Writing ', resdir+img.imagename+'_pyrank.fits'))
```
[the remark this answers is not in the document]
darafferty: Done.

test/compare_outputs.py
darafferty: Good suggestion -- done.

test/compare_outputs.py
darafferty: Done.

darafferty:
I think all comments have been addressed. I also reduced the verbosity of the comparison as requested.

gmloose (approved):
Looks good to me.

PR #238, opened 2024-09-05
Add macos arm binary wheels

gmloose:
Whenever a new release tag is created. the CI/CD pipeline now creates the following artifacts and publishes these on PyPI:

* a source distribution as tar archive
* x86-64 binary wheels for Linux
* x86-64 binary wheels for MacOS 12
* arm64 binary wheels for MacOS 14

Closes RAP-739


tammojan (approved):
Works on my mac!

PR #239, opened 2024-09-06
Update changelog

gmloose:
Update the changelog. This should have been done _before_ v1.11.0 was released, but it was overlooked, unfortunately. TODO: Find a way to automate this.

### DIFF
--- a/bdsf/_changelog.py
+++ b/bdsf/_changelog.py
@@ -8,6 +8,39 @@ def changelog():
     """
     PyBDSF Changelog.
     -----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


gmloose:
One way to get more or less usable changelog entries is with the following command:
```
git log --pretty=format:"    %ad - %s%n" --date=format:"%Y/%m/%d" v1.10.3..v1.11.0
```
Where you need to change the version ranges to suit your needs, of course.

darafferty (approved):
Looks good. The `git log` command is a nice shortcut that I wasn't aware of.

PR #240, opened 2024-10-09
Fix bug that resulted in ncores being at most 8

darafferty:
This PR fixes a bug in the multiprocessing module that resulted in the number of cores used in processing being limited to 8. It also improves the method used to get the number of available cores (following https://git.astron.nl/RD/rapthor/-/merge_requests/170) and fixes some minor formatting issues.

### DIFF
--- a/bdsf/multi_proc.py
+++ b/bdsf/multi_proc.py
@@ -11,28 +11,27 @@ Adapted from a module by Brian Refsdal at SAO, available at AstroPython
 from __future__ import print_function
 import traceback
 import sys
-
+import os
 import numpy
-_multi = False
+import multiprocessing
 _ncpus = 1
 
-try:
-    # May raise ImportError
-    import multiprocessing
-
-    # Set spawn method to "fork". This is needed for macOS on Python 3.8+ where the
-    # default has been changed to "spawn", causing problems (see the discussion at
-    # https://github.com/ipython/ipython/issues/12396)
-    if sys.platform == 'darwin':
-        if sys.version_info[0] == 3 and sys.version_info[1] >= 8:
-            multiprocessing.set_start_method('fork')
-    _multi = True
-
-    # May raise NotImplementedError
-    _ncpus = min(multiprocessing.cpu_count(), 8)
-
-except:
-    pass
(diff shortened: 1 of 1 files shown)


bdsf/multi_proc.py
```diff
@@ -11,28 +11,27 @@
...
+#
+# Note: since macOS (Darwin) does not support os.sched_getaffinity(), we use
+# multiprocessing.cpu_count() instead
+if sys.platform == 'darwin':
+    if sys.version_info[0] == 3 and sys.version_info[1] >= 8:
+        # We need to set spawn method to "fork" for macOS on Python 3.8+ where
+        # the default has been changed to "spawn", causing problems (see the
+        # discussion at https://github.com/ipython/ipython/issues/12396)
+        multiprocessing.set_start_method('fork')
+    _ncpus = multiprocessing.cpu_count()
```
gmloose, line 32: This will still give you too many processes on MacOS. I'm not a 100% sure if `os.sched_getaffinity` is available on MacOS. The documentation is silent on that. We could consider to use `len(psutil.Process().cpu_affinity())` as a more portable solution, however, maybe that function is also not supported on MacOS, and we would drag in a dependency on another package (`pustil`) for just a single command.

gmloose (commented):
Good catch, the (ab)use of `min` in this case 🤔

bdsf/multi_proc.py
tammojan: `os.sched_getaffinity` is not available on my mac, python 3.9.13 nor 3.12.5. On MacOS, also `psutil.Process()` does not have a method `cpu_affinity` on MacOS.

gmloose:
This morning I realized that your is not that obvious after all, and I wonder if it will work. I was put on the wrong leg by your sentence that due to a bug the maximum number of core being used was limited to 8. However, you use the `min()` function. So, only in cases where `multiprocessing.cpu_count()` returns less than 8, would it return 8. Hence, the real question is: is `multiprocessing.cpu_count()` broken?

gmloose (approved):
LGTM.

Issue #176, opened 2022-02-18
logging when using bdsf in other codes


revoltek:
When using pybdsf within other codes that also make use of  the "logging" module some of the logs are messed up. What I noticed is that some of the logs of the main code is now internally handled by bdsf and e.g. redirected to the log file of bdsf.

rvweeren:
I have the same experience, PyBDSF takes over my own logging if I use the module in my code. Can this be fixed?

Issue #233, opened 2024-06-25
mylogger.py removes all existing logging handlers


zack466:
I noticed that in `mylogger.py`, when `init_logger()` is called, it removes all existing handlers from the root logger ([source](https://github.com/lofar-astron/PyBDSF/blob/cc79ebab2f00d203741731efde5704ceb766dfb6/bdsf/mylogger.py#L25)). However, this means that when pybdsf is called from a scripting context, it deletes all of the existing logging handlers set up by the calling code, completely eliminating all non-pybdsf logging logic. This is very unexpected and definitely should not happen. I think it would be helpful to assign constant names to the pybdsf-specific logging handlers and only remove those named handlers from the root logger in `init_logger()` rather than all of them.

PR #245, opened 2024-11-15
Fix unintended removal of handlers from the root logger

darafferty:
This PR fixes issues #176 and #233. Some cleanup was also done to remove unneeded files and references to the old "PyBDSM" name.


gmloose (commented):
There are still a number of references to `bdsm` or `pybdsm` (with different capitalizations), but let's not pollute this MR with cleaning up these. Except maybe for one: remove line 65 in `pyproject.toml`:
```
pybdsm = "bdsf.pybdsf:main [ishell]"
```
Or do we still want to retain backward compatibility by supporting the name `pybdsm` as executable script?

darafferty:
Yes, good point. Indeed, I don't think we need to make the `pybdsm` script any longer, as surely it's been long enough. I removed the line from `pyproject.toml`.

gmloose (approved):

PR #248, opened 2024-11-25
Update changelog and whats_new

gmloose:
Updated Change Log and What's New


gmloose:
I considered removing Python 3.8 support, but decided that I should maybe wait for the next release to do so. @darafferty what do you think?

darafferty:
Yes, agreed -- from my point of view, there's no pressing reason to drop Python 3.8. (That said, it seems the oldest version I'm currently using is 3.9.)

PR #251, opened 2025-04-03
Issues/247 support numpy2.0

### LINKED ISSUE #247: Support for NumPy 2.0
Hello Team, 

The VLASS calibration pipeline uses PyBDSF during the imaging stage. Currently, the calibration team is working to make the pipeline compatible with NumPy 2.0. 

Do you plan to support NumPy 2.0 in the near future? If so, could you please provide some insights into the timeline?

Thank you,
Ajay

gmloose:
Updated the code base to properly work with the new NumPy 2.x API. 

This pull request resolves #247.


test/reference_outputs/misc/parameters_used
```diff
@@ -83,7 +83,7 @@ multichan_opts                          False
 ncores                                  2
 opdir_overwrite                         'overwrite'
 outdir                                  None
-    Value used                          '/home/runner/work/PyBDSF/PyBDSF/test'
+    Value used                          '/home/marcel/code/PyBDSF/test'
```
tammojan, line 86: This seems specific to your system.

test/reference_outputs/catalogues/tbdsf_process_image.in_fits.pybdsf.srl.ds9.reg
tammojan, line 1: Any idea why do the numbers change? At first sight, the change is very small, so I think acceptable if we have an idea about the reason.

.github/workflows/ci.yml
```diff
@@ -1,21 +1,51 @@
...
+            toml = tomllib.load(f)
+
+        # Extract key-value pairs from the environment used by cibuildwheel
+        matches = re.findall(
+            r'(\w+)=(".*?"|\S+)',
+            toml["tool"]["cibuildwheel"]["environment"]
+        )
+
+        # Add every key-value pair to the environment, so that if a value contains an
+        # environemt variable, it can be expanded before it is added to GITHUB_ENV.
```
tammojan, line 32:
```suggestion
        # environment variable, it can be expanded before it is added to GITHUB_ENV.
```

tammojan (commented):
This is too much for an in-depth review, but I heard that you tested with numpy 1 as well. Since much of the changes are renames, they should fall through at compile-time, so I believe that they work.

test/reference_outputs/misc/parameters_used
gmloose: Yep, I can revert it, but it's part of a test that's normally not run in CI.

tammojan (approved):

darafferty (approved):
Wow, nice job! Everything looks good to me (and it's nice to have confirmation that the output is effectively the same).

olebole:
Thank you very much! Will you tag a new version, or shall I take this as a patch for Debian?

tammojan:
Is tomorrow early enough for a tag?

gmloose:
I will create a new (tagged) release today.

olebole:
@gmloose thank you very much! I just uploaded the updated package to Debian, and [the builds (including tests)](https://buildd.debian.org/status/logs.php?pkg=pybdsf&ver=1.13.0-1) are fine so far, so this will hopefully migrate to Debian testing before our freeze.

gmloose:
@olebole, happy to hear that everything works fine. I found one minor glitch in our deployment pipeline (I need to upgrade the Intel-based MacOS runner), but that shouldn't have any impact on Debian. I hope indeed we're still in time for the Debian freeze; it would be a pity if it were dropped.

PR #252, opened 2025-04-07
Fix show_fit command

darafferty:
This PR fixes the `show_fit()` command of the interactive shell for Python 13.


gmloose (approved):
I can confirm that I can now also run `pybdsf` and call the `show_fit` function successfully inside a Python 3.13 docker container.

PR #267, opened 2025-10-30
Fix binary wheel build for Python 3.14

gmloose:
In Python 3.14, the default method that `multiprocessing` uses to start a process on POSIX platforms changed from `fork` to `forkserver`. We cannot use `forkserver`, unless we restructure our code, so we need to explicitly set the start method to `fork` when running on Linux using Python 3.14 or later.

Fixes #265


darafferty (approved):
Looks good -- nice that the fix was so simple.

darafferty (approved):
Looks good.

PR #272, opened 2026-04-28
Speedup `gaus_2d()`

AlexKurek:
On a 127 MB, 5762*5762 image, `atrous_do = True` total running time:

```
before: 0:2:30
after:  0:3:3
```
22% faster.

`NGC6048_9-MFS-image.pybdsf.gaul` files are identical.


AlexKurek:
Im sorry, this is not the one I was going to post.

PR #278, opened 2026-05-19
Fix imports in `rmsimage.py`

AlexKurek:
Removed:

- multiple imports of numpy and scipy.ndimage at the top of the file
- obsolete imports from Python 2
- local imports `from math import sqrt` from within the `check_rmsmap` and `check_meanmap` functions. Instead, a global import `from math import sqrt, log` was added at the top of the file

### DIFF
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -13,8 +13,7 @@ import itertools
 from concurrent.futures import ThreadPoolExecutor
 
 import numpy as np
-import scipy.ndimage as nd
-from scipy import interpolate
+from scipy import interpolate, ndimage
 
 from .image import Op, Image, NArray, List
 from . import const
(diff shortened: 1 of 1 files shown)


bdsf/rmsimage.py
```diff
@@ -47,17 +40,17 @@ def mapcoord_threaded(a, axs, *args, ncores=8, **kwargs):
     # _interpolation.py in scipy
     order = kwargs.get("order", 3)
     if order > 1:
-        a = ndimage.spline_filter(a, order,
-                                  output=numpy.float64,
+        a = nd.spline_filter(a, order,
```
gmloose: I would prefer the use of `ndimage`. Contrary to `np` being the ubiquitous abbreviation for `numpy`, I don't consider `nd` to be that for `scipy.ndimage`. So, I suggest that you undo the change here (and in a few other lines) and change line 16 (and merge it with 17) like:
```
from scipy import interpolate, ndimage
```

gmloose (changes requested):
Thanks for cleaning up the scattered imports. I have one request for change.

bdsf/rmsimage.py
AlexKurek: Good point, it looks better immediately. Done in https://github.com/lofar-astron/PyBDSF/pull/278/commits/5723ab16af67547f42edb5651eddd82ad617f0da.

PR #279, opened 2026-05-19
Remove dead code from `ndimage_alg`

AlexKurek:
This is a dead code, its not used or returned by the function.


gmloose:
I wonder if it's dead code, or (worse) a missing assignment like:
```
            img.pyrank = pyrank
```
As in [line 138](https://github.com/lofar-astron/PyBDSF/blob/dd3500c1db6be22bd4586e018c77e87a944f700a/bdsf/islands.py#L138). 
@darafferty?

darafferty:
Hmm, good question... I'm not sure why the code was left, but perhaps it happened when the option to use user-supplied source locations was added. Anyway, I think it's fine to remove.

AlexKurek:
This was duplicated, Im sorry. Done in https://github.com/lofar-astron/PyBDSF/commit/07af214cd2d87418bfa6133a820d165b738c6458.

PR #280, opened 2026-05-21
Remove `ntot_wvgaus` as a dead code

AlexKurek:
As this variable is never modified, line 304 serves no purpose.


gmloose (approved):
Good catch.

PR #281, opened 2026-05-21
Fix renumbering of islands

darafferty:
This PR fixes an issue with renumbering of islands done when merging islands in the wavelet module, causing incorrect or missing island merges.


AlexKurek:
I have also noticed this, but I wasn't sure because it makes a lot of differences to the .gaul file.
So, the reference results at the CI probably need updating.

darafferty:
Indeed, the CI now fails on a couple of images (I think those that are sensitive to how Gaussians are assigned to the islands). I will update the reference files.

gmloose (approved):
That was a subtle bug 😄. I'll take the changed reference outputs for granted.

PR #282, opened 2026-05-21
Make `pixdist2angdist` faster

AlexKurek:
For 1.5 GB FITS the function is >2x faster.

before:
`   400452    2.451    0.000   21.837    0.000 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/readimage.py:585(pixdist2angdist)`
after:
`   400452    2.103    0.000    9.663    0.000 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/readimage.py:585(pixdist2angdist)`
.gaul files are identical.

Regarding "Trigonometry and bitwise identity" in the comments: I tried without this and there were numerical differences at 4. decimal place and beyond.


gmloose (commented):
Looks good, though it's unclear to me how much speed improvement this will give for the whole (not just this function). Maybe @darafferty would like to double check?

AlexKurek:
These are the results from cProfile, so:
`   ncalls  tottime  percall  cumtime  percall filename:lineno(function)`
21.837 - 9.663 = ~12 seconds for 1.5 GB FITS file on Xeon 6238R CPU. This is not much, but I assume that much larger files are being processed in surveys etc.

AlexKurek:
...let me explain a bit more on this.
`ncalls` shows, that this function is called 400k times, so vectorising the calls could be beneficial. However, this only makes sense when the function itself has been optimised as much as possible. Therefore, optimising a helper function can open the way to optimising the function that calls it.
`tottime` shows how much time is spent withing the function, not including numpy, scipy etc.
`cumtime` shows a total time the function takes in a PyBDSF run.

You can also use `line_profiler` (output in attachemnt):
`import line_profiler`
decorator for the function you want to profile:
`@line_profiler.profile`
Run:
`LINE_PROFILE=1 python run_pybdsf.py`
See output:
`python -m line_profiler -rtmz profile_output.lprof`

There is also `memory_profiler `:
```
from memory_profiler import profile
@profile # decorator
python -m memory_profiler --include-children run_pybdsf.py
```


line_profiler output: [profile_output_2026-05-21T102741.txt](https://github.com/user-attachments/files/28135330/profile_output_2026-05-21T102741.txt)

Another thing - results from snakeviz (`snakeviz out.prof`, where out.prof is the output from Cprofile) shows the relations between the functions:
Screenshot 2026-05-22 at 07-47-30 out

bdsf/readimage.py
gmloose, line 1: I'm not really a WCS guru. The first part of the modifications (up to line 616) make perfect sense to me. The rest is probably OK, but I'd have to ask @darafferty.

One other thing. There's also an `angdist2pixdist()` method. Shouldn't that be optimized in a similar way?

bdsf/readimage.py
```diff
@@ -582,27 +582,70 @@ def angdist2pixdist(self, img, angdist, pa, location=None):
...
+            xy_arrN[1, 1] = y2
+
+            # Vectorization and bypassing .tolist() overhead
+            # Instead of invoking the slow `img.pix2sky` wrapper element by element, 
+            # pass both coordinates to the WCS in a single, vectorized call
+            if hasattr(wcs_obj, 'wcs_pix2world'):
+                sky = wcs_obj.wcs_pix2world(xy_arrN, 0)
+            else:
+                # Fallback for older pywcs versions
+                sky = wcs_obj.wcs_pix2sky(xy_arrN, 0)
```
gmloose: Is this still necessary. Unless I'm mistaken, `wcs_pix2sky` was replaced by `wcs_pix2world` in `astropy` version 0.1!

bdsf/readimage.py
```diff
@@ -582,27 +582,70 @@ def angdist2pixdist(self, img, angdist, pa, location=None):
             result = N.mean(img.wcs_obj.acdelt[0:2])
         return result
 
+
     def pixdist2angdist(self, img, pixdist, pa, location=None):
-        """Returns the angular distance in degrees for a given distance in pixels
+            """Returns the angular distance in degrees for a given distance in pixels
```
gmloose: Something went wrong with the indentation. This whole method has 8 spaces indentation, instead of 4.

gmloose (commented):
Some minor comments.

bdsf/readimage.py
AlexKurek: > There's also an angdist2pixdist() method. Shouldn't that be optimized in a similar way?

I dont see the other method in cProfile, probably its not executed at all.

darafferty (approved):
Looks good to me. I think the argument for optimizing the function makes sense.

PR #286, opened 2026-05-30
Make statusbar faster

AlexKurek:
This does not require an extensive testing, since I havent touched computations. Just look at `ncalls` and it explains everything. Now printing and checking the terminal size occurs only 5x / sec.
For a 1.5 GB FITS:
before:
```
     5823    0.018    0.000   10.835    0.002 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/statusbar.py:70(increment)
     5841    0.036    0.000   10.809    0.002 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/statusbar.py:46(__print)
     5851    0.023    0.000   10.764    0.002 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/statusbar.py:33(__getsize)
```
after:
```
     5823    0.007    0.000    0.208    0.000 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/statusbar.py:79(increment)
       87    0.001    0.000    0.260    0.003 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/statusbar.py:56(__print)
       97    0.001    0.000    0.294    0.003 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/statusbar.py:40(__getsize)
```
This 3 functions are 52x faster.
The entire run is 2.6% faster.

### DIFF
--- a/bdsf/statusbar.py
+++ b/bdsf/statusbar.py
@@ -58,12 +58,15 @@ class StatusBar():
 
         sys.stdout.write('\x1b[1G')
 
-        # Check to prevent ZeroDivisionError if there are no items to process
+        # Handle the case where there are no items to process (prevents from
+        # dividing by zero later)
         if self.max == 0:
             sys.stdout.write(self.color + self.text + '[] 0/0\033[0m\n')
         else:
-            # Recalculate comp here (moved from increment) immediately after __getsize()
-            # to ensure the progress bar adapts correctly if the terminal was resized
+            # We are about to divide by self.max. We are inside the 'else' block,
+            # so self.max is not 0, meaning this division will not crash.
+            # We calculate this right after checking the terminal size (__getsize) 
+            # so the progress bar adapts to window resizing.
             self.comp = int(float(self.pos) / self.max * self.columns)
             sys.stdout.write(self.color + self.text + '[' + '=' * self.comp + self.busy_char + '-'*(self.columns - self.comp - 1) + '] ' + str(self.pos) + '/' + str(self.max) + '\033[0m')
             sys.stdout.write('\x1b[' + str(self.comp + 2 + 44) + 'G')
--- a/bdsf/statusbar.py
+++ b/bdsf/statusbar.py
@@ -60,7 +60,7 @@ class StatusBar():
 
         # Handle the case where there are no items to process (prevents from
         # dividing by zero later)
-        if self.max == 0:
+        if self.max <= 0:
             sys.stdout.write(self.color + self.text + '[] 0/0\033[0m\n')
         else:
             # We are about to divide by self.max, but it is safe in the 'else' block.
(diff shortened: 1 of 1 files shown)


bdsf/statusbar.py
```diff
@@ -35,21 +40,28 @@ def __getsize(self):
...
     # redraw progress bar
     def __print(self):
+        # Update terminal size with each actual frame rendering
         self.__getsize()
 
         sys.stdout.write('\x1b[1G')
         if self.max == 0:
             sys.stdout.write(self.color + self.text + '[] 0/0\033[0m\n')
         else:
+            # Recalculate comp in case the terminal size (self.columns) has changed
```
gmloose: Comment doesn't match with the conditional (`self.max != 0`). So either the comment or the code is incorrect. Also, it's not completely clear for me why you moved this line from the `increment()` method into the `__print()` method.

gmloose (commented):
I think it's OK to only update every 200 ms. I'm don't completely understand the other changes that I commented on.

bdsf/statusbar.py
AlexKurek: We want the progress bar to handle terminal resizing at any moment, so the calculation of `self.comp` must always happen immediately after checking the terminal size.
If/else is there to prevent division by zero error.

gmloose: I understand that you want to prevent division by zero errors. My point is that the comment says that `self.columns` has changed, whereas the actual `if`-statement checks `self.max == 0`. That doesn't rhyme.

AlexKurek: OK, I see what you mean now. I have refined the comments, is it better now?
EDIT: Im sorry, I have titled a commit `Refine comments # 2` (without spacebar) and it linked to PR number 2. I have reverted this commit and pushed a new one.

gmloose: The only thing you don't protect against is if `self.max` is less than zero. But I see that the original code also does `if self.max == 0`. That's probably fine, but curiously enough `__init__()` contains a `if self.max > 0`. 🤔

AlexKurek: Maybe in the `__print` method we should change `if self.max == 0:` to `if self.max <= 0:`?

gmloose: Guess it wouldn't hurt, and maybe even catches some rare corner cases that might otherwise trigger weird behavior. OTOH, I just looked what happens if you do something like `"=" * -3` and Python will simply return an empty string.

AlexKurek: Done in https://github.com/lofar-astron/PyBDSF/pull/286/commits/f6350d787d529ee9bcf8dcc7afed803203943991.

PR #287, opened 2026-05-30
Speedup `getTerminalSize`

AlexKurek:
Python 3.3+ has a build in `get_terminal_size` in `shutil`, so this code is not needed.
Similar as https://github.com/lofar-astron/PyBDSF/pull/286, it does not require extensive testing, since I havent touched computations.

For a 1.5 GB FITS:
before:
`     5851    0.096    0.000   16.327    0.003 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/functions.py:1970(getTerminalSize)`
after:
`     5851    0.006    0.000    0.070    0.000 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/functions.py:1971(getTerminalSize)`
233x faster.

### DIFF
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -1,6 +1,7 @@
 # some functions
 from __future__ import print_function
 from __future__ import absolute_import
+from shutil import get_terminal_size
 
 try:
     # For Python 2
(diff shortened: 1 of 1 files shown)


bdsf/functions.py
```diff
@@ -1968,40 +1969,18 @@ def make_src_mask(mask_size, posn_pix, aperture_pix):
...
-    try:
-        return tuple(int(os.getenv(var)) for var in ("LINES", "COLUMNS"))
-    except:
-        pass
-    # Give up. return 0.
-    return (0, 0)
+
+    # shutil.get_terminal_size accepts a fallback in the format (columns, lines).
+    # We set it to (80, 25) to maintain the identical default values as the old code.
+    size = get_terminal_size(fallback=(80, 25))
```
gmloose: Where did you find those defaults? I don't see them in the original code. In will return `(0, 0)` if it cannot determine the terminal size. I can live with `(80, 25)`, but also with `(80, 24)`, which is the default if you don't provide a `fallback`. However, I'm not 100% sure if `(0, 0)` is treated as special value elsewhere in the code.

gmloose (commented):
Nice improvement, but I have a question about the fallback value you're using (see above).

bdsf/functions.py
AlexKurek: (80, 25) is the default terminal size. But I guess I shouldn't have changed that in this PR, since this PR is about speed optimization. For me both (80, 25) and (0, 0) worked, but I have restored (0, 0) in this PR.

gmloose: Thanks, looks good.

PR #289, opened 2026-06-01
Cleanup of `rmsimage.py`

AlexKurek:
- `max_isl_brightsize` is never used.
- `mylog = mylogger.logging.getLogger("PyBDSF."+img.log+"Rmsimage.Calcmaps ")` was a duplicate - the same line is 2 lines above.
- `thratio`, `avg_max_isl_size`, `maxrms`, `maxmean`, `pt_src_scale` are computed but not used.

### DIFF
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -162,7 +162,6 @@ class Op_rmsimage(Op):
         # the bright source is embedded inside a large island or not. If it is,
         # exclude it from the bright-island list. Also find the size of the
         # largest island at this threshold to set the large-scale rms_box
-        bright_threshold = threshold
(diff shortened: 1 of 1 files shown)


AlexKurek:
- after removing the above variables also `max_isl_size_lowthresh`, `bright_threshold`, `isl_snr`, `mask_small` and `isl_size_lowthresh` are not needed
- `from . import _cbdsm` is not used
- `from .image import Image, NArray, List` is not used
- moved `floor, ceil` import tothe top of the file

Changing the name of the PR, since its not longer so minor.

darafferty:
Looks good to me — nice to have all this cruft removed!

PR #290, opened 2026-06-03
Fix brightest source estimation

### LINKED ISSUE #288: `rmsimage.py`: the size of the brightest source is always estimated using the fallback formula
There is a try / except block:
https://github.com/lofar-astron/PyBDSF/blob/master/bdsf/rmsimage.py#L121
```
        try:
            brightsize = int(round(2.*img.beam[0]/cdelt[0]/fwsig*
                               sqrt(2.*log(img.max_value/(kappa1*crms)))))
        except:
            brightsize = int(round(2.*img.beam[0]/cdelt[0]/fwsig))
```

but `crms` is defined below this code block in line 132, so the exception always occurs. Is it OK to bring
```
        crms = crmss[0]
```
above this try / except block?

AlexKurek:
Fixes https://github.com/lofar-astron/PyBDSF/issues/288.

### DIFF
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -116,20 +116,29 @@ class Op_rmsimage(Op):
         img.use_rms_map = None
         img.mean_map_type = None
 
+        crms = crmss[0]
         # 'size' of brightest source
         kappa1 = 3.0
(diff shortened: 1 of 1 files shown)


darafferty:
Looks good — nice fix and comments!

PR #291, opened 2026-06-03
Fix initial Gaussian size estimation in `inigaus_nobeam`

AlexKurek:
If Im not mistaken, instead of measuring the current mask, the code sums pixels from all masks at once. Every component gets an identical, artificially inflated initial size.


darafferty:
Looks good, well spotted.

PR #293, opened 2026-06-05
Remove dead code from `islands.py`

AlexKurek:
This is a dead code computing large matrix in a loop.


darafferty:
Thanks — another good find! PyBDSF's had a messy development history, so it's great to clean up some of the mess.

PR #294, opened 2026-06-05
Fix computing the number of unmasked pixels assigned to island

AlexKurek:
In the constuctor there is
`copy=False`
So `if not copy:` block will always be executed. Within this block:
`isl_mask = (labels[tuple(bbox)] == idx)`
so True means pixels within the island.
This mask is inverted below:
`N.logical_not(isl_mask, isl_mask)`
Since the mask is inverted, currently this line:
`pixels_in_isl = N.sum(~N.isnan(self.image[self.mask_active]))`
seems to count not-NaN pixels *outside* the island. To fix it:
`pixels_in_isl = N.sum(~N.isnan(self.image[~self.mask_active]))`

Am I right? This produces chenges in I think every line in the .gaul file in `Isl_Total_flux` column.


darafferty:
Yes, you're right. I noticed this myself when looking at one of the later PRs (having forgotten seeing the fix here). I'll merge this one first and then look at the later ones in detail.

PR #295, opened 2026-06-06
Fix `total_flux` computation

AlexKurek:
Background noise within the bounding box (but outside the island) was included in the island's total flux. To fix this, we mask the area outside the island.
There are a lot of differences in the .gaul file in `Isl_Total_flux` after this.


AlexKurek:
This is how `self.image[in_bbox_and_unmasked]` currently looks. It confirms, that total flux of the bounding box is computed, not the total flux of a source.

Adding this:

```
Its primary use is a container for all kinds of data describing island.
    """
    _img_save_count = 0

    def __init__(self, img, mask, mean, rms, labels, bbox, idx,
                 beamarea, origin=None, noise_mask=None, copy=False):
```
and this:
```
        self.islmean = bbox_mean_im[in_bbox_and_unmasked].mean()

        if Island._img_save_count < 5:
            import matplotlib.pyplot as plt
            plot_data = N.full(self.image.shape, N.nan)
            plot_data[in_bbox_and_unmasked] = self.image[in_bbox_and_unmasked]
            
            plt.figure(figsize=(6, 5))
            plt.imshow(plot_data, origin='lower', cmap='viridis')
            plt.colorbar(label='Flux')
            plt.title(f'Island ID: {idx}')

            plt.savefig(f'island_{Island._img_save_count}.png', bbox_inches='tight')
            plt.close()
            
            Island._img_save_count += 1

        self.total_flux = N.nansum(self.image[in_bbox_and_unmasked])/beamarea
```
will show:
island_0
island_1
island_2
island_3
island_4

AlexKurek:
After this PR:
island_0
island_1
island_2
island_3
island_4

Tagging @gmloose since you may be interested to see this.

AlexKurek:
Figure_1
This is what I got for a 20k x 20k test image.

darafferty:
Very nice analysis! I agree with your conclusion that the mask was not being considered, so the fluxes were not calculated as intended.

AlexKurek:
I have resolved the conflict after recent merge of https://github.com/lofar-astron/PyBDSF/pull/294. Its seems that the reference files in CI need to be updated.

darafferty:
Yes, I noticed the CI fails now, but interestingly I can't reproduce it on my machine — the new images I generate are basically identical to the reference ones. I guess we could relax the tolerance for the checks again (currently it's 0.01, and I think increasing it to 0.02 would work). @gmloose, what do you think?

AlexKurek:
Perhaps CI fails because the differences from previous merged PRs have accumulated, so essentially not because of this PR? This PR will not change the images, only the catalog.
In that case there would be no need for relaxing the tolerance.

darafferty:
Yes, I think you're right about the reason for the CI failure, as it first occurred I think when I merged #339 yesterday, but I can't get the test script to fail on my machine either on the current master or on this branch. It's strange.

AlexKurek:
Yes, it's strange. It's also not the first time: this shouldn't have caused the CI to crash either: https://github.com/lofar-astron/PyBDSF/pull/324.

AlexKurek:
There is something wrong about CI. I have updated the branch in https://github.com/lofar-astron/PyBDSF/pull/324 to force the CI to re-run the test and it passed.
My suggestion would be to merge some other PR first, to that here the branch can also be updated, and we will se if the test will work also here.

darafferty:
OK, lets try that and see how it goes.

AlexKurek:
It worked - the test is passed.

darafferty:
OK, this MR looks good. I'll merge it, and then there is still the one regarding the error on the flux (#301 I think).

AlexKurek:
I have fixed the conflicts in https://github.com/lofar-astron/PyBDSF/pull/301.

PR #296, opened 2026-06-06
Fix computing brightest pixel in the island

AlexKurek:
In the bounding box: if there is a brighter pixel outside the island than inside - it will be selected as the brightest in the island. Fix it by masking the area of the bounding box that do not belong to the island.
This may be very rare situation, since eg. for my 1.5 GB test image the .gaul files are identical.


darafferty:
Nice find! I expect you're right and it is rare, but I can imagine it could happen for some complex sources that are broken into multiple islands.

PR #298, opened 2026-06-09
Remove `getTerminalSize`

AlexKurek:
Currently this function makes no sense, it only refers to `shutil.get_terminal_size`. Better to call `get_terminal_size` directly. its working fine.


gmloose (approved):
LGTM.

PR #299, opened 2026-06-10
Fix statistics of the residual image

AlexKurek:
First:
`resid_gaus = cp(img.ch0_arr)`

Later the algorithm checks if the mask is defined. If it is, it injects "new" NaNs into `resid_gaus`
```
if isinstance(img.mask_arr, N.ndarray):
    pix_masked = N.where(img.mask_arr == True)
    model_gaus[pix_masked] = N.nan
    resid_gaus[pix_masked] = N.nan (1)
```

The algorithm tries to isolate the non-Nan pixels:
`resid_gaus_non_masked = resid_gaus[N.where(~N.isnan(img.ch0_arr))`
The problem is, that it searches for the pixels that are not NaN in the original image. But "other" NaNs were added to `resid_gaus` in `(1)`. So `resid_gaus_non_masked` still may have NaNs. In that case later `N.mean()`, `N.std()` etc. will return NaN-s for the entire array.

EDIT:
Reverted [Same for shapelets](https://github.com/lofar-astron/PyBDSF/pull/299/commits/6ef65ca8bd18098bea24a6d403a1470663286b18) since it crashes the test. Maybe lets stick to `resid_gaus` in this PR.


AlexKurek:
The test is failing for unrelated reasons ( https://github.com/lofar-astron/PyBDSF/pull/309#issuecomment-4717433470 )

gmloose (approved):
LGTM. It's indeed weird that a similar change for `resid_shap` causes an error. I'm not sure what triggers this, cause I don't see how `numpy.where()` could return something that cannot be used as indices into an array.

PR #300, opened 2026-06-10
Prepare for release 1.14

gmloose:
Update Changelog for release 1.14


gmloose:
I think it's time to make a new release for at least two reasons:

1. The previous release is more than one year old
2. There's no binary wheel for Python 3.14 for that release. Python 3.14 ships with Ubuntu 26.04, which is the new Ubuntu LTS version.

Do I need to change anything else, besides the Changelog?

darafferty:
I think updating the changelog is all that's needed. Looks good!

darafferty (approved):

PR #301, opened 2026-06-11
Fix `total_fluxE` computation

AlexKurek:
`total_fluxE` was computed on the entire bounding box. Fix it by computing only on the island area.
There are differences in the .gaul file in every line at one of the columns, usually not large.


darafferty:
Looks good!

PR #304, opened 2026-06-11
Shapelets: fix indentation in `getzeroes_matrix`

AlexKurek:
Currently
`for i in range(cf.shape[0]):`
is actually not doing anything and the entire if block:
`if npts > 3 and not N.isnan(cf[i,cen]):`
is executed only for the last row in the matrix.

### DIFF
--- a/bdsf/shapelets.py
+++ b/bdsf/shapelets.py
@@ -216,35 +216,35 @@ def getzeroes_matrix(mask, cf, cen, cenx):
         npts = len(l)-sum(l)
 
         #print 'npts = ',npts
-    if npts > 3 and not N.isnan(cf[i,cen]):
-        mrow=mask[i,:]
-        if sum(l) == 0:
-            low=0
-            up=cf.shape[1]-1
-        else:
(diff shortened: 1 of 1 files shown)


darafferty:
Nice find — I guess I'm not too surprised this went unnoticed for as long as it did, as I don't think the shapelets module is used much. Thanks for the fix!

AlexKurek:
https://github.com/AlexKurek/PyBDSF/blob/bdbfdfcba53f51a27e42cc3cec6d58e35b306968/bdsf/shapelets.py#L325
By the way, what do you think about this? The same variable is defined two lines above.

darafferty:
Hmm, good question. It looks like a temporary fix or test was made to `beta_arr`, but it was never finalized... I would guess that the more sophisticated method of making the array of beta to test was not really working properly, so the simple hard-coded array was put in instead, but I don't know for sure (this was all done before my time).

AlexKurek:
I have the same impression. Maybe somebody just worked out that this strictly given range works better.

darafferty:
Could be indeed. It would be good to do some testing to understand it better, but anyway for now I'll go ahead and merge this PR.

PR #305, opened 2026-06-12
Use fork-style multiprocessing context

gmloose:
PyBDSF currently relies on fork-style multiprocessing. The spawn and forkserver start methods are not supported because parts of the codebase assume inherited interpreter state and are not safe for re-import during process startup.


darafferty (approved):
Nice fix!

PR #306, opened 2026-06-12
Make printing of stats of residual images optional

bnikolic:
Calculating them has a significant computational load for my SKA reference run

### DIFF
--- a/bdsf/make_residimage.py
+++ b/bdsf/make_residimage.py
@@ -155,17 +155,18 @@ class Op_make_residimage(Op):
                 resid = resid_shap[tuple(isl.bbox)]
                 self.calc_resid_mean_rms(isl, resid, type='shap')
 
-            # Calculate some statistics for the Shapelet residual image
-            non_masked = N.where(~N.isnan(img.ch0_arr))
-            mean = N.mean(resid_shap[non_masked], axis=None)
-            std_dev = N.std(resid_shap[non_masked], axis=None)
-            skew = stats.skew(resid_shap[non_masked], axis=None)
-            kurt = stats.kurtosis(resid_shap[non_masked], axis=None)
-            mylog.info("Statistics of the Shapelet residual image:")
-            mylog.info("        mean: %.3e (Jy/beam)" % mean)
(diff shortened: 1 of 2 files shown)


bdsf/opts.py
```diff
@@ -710,6 +710,11 @@ class Opts(object):
                                  "faster but also uses much more memory.",
                              group = "atrous_do")
 
+    residual_stats_do = Bool(False,
+                             doc = "Calculate and print to log statistics of residual "\
+                             "images (there is a computational cost)",
+                             group = "atrous_do")
```
darafferty: The new option's `group` value means that it  is grouped under the atrous options, where it doesn't really fit. I think it would be better to make it part of the output options (so `group = "output_opts"`). Please also move it to the output options section, which starts around line 867.

bdsf/make_residimage.py
```diff
@@ -85,18 +85,19 @@ def __call__(self, img):
...
-        std_dev = N.std(resid_gaus_non_masked, axis=None)
-        skew = stats.skew(resid_gaus_non_masked, axis=None)
-        kurt = stats.kurtosis(resid_gaus_non_masked, axis=None)
-        stat_msg = "Statistics of the Gaussian residual image:\n"
-        stat_msg += "        mean: %.3e (Jy/beam)\n" % mean
-        stat_msg += "    std. dev: %.3e (Jy/beam)\n" % std_dev
-        stat_msg += "        skew: %.3f\n" % skew
-        stat_msg += "    kurtosis: %.3f" % kurt
-        mylog.info(stat_msg)
+        if img.opts.residual_stats_do:
```
darafferty, line 88: This section is for the Gaussian residual image, but the statistics are also calculated for the shapelet one as well, when done (starting around line 158). Please make the option apply to both.

darafferty (commented):
Thanks for the PR — I can see that these calculations could be costly in some cases, so good to have an option to turn them off.

bnikolic:
Thanks for your comments -- I've updated and pushed the changes.

darafferty:
Looks good!

PR #308, opened 2026-06-13
Fix typo in `shapelet_check_centre`

AlexKurek:
The logic of this function currently makes no sense; this is an obvious typo.


darafferty (approved):
Thanks, another good catch!

PR #312, opened 2026-06-16
Clean `shapelets.py`

AlexKurek:
There may be more bugs in this module, but first it would be good to clean it up. There are unused and redundant imports and unused variables.


gmloose:
LGTM. Thanks for taking the time to clean up stuff.

PR #314, opened 2026-06-16
Update reference output files

darafferty:
This PR updates the reference files used for the CI tests. The files changed due to the fixes applied in #309.


gmloose (approved):
Since you approved #309, I guess it makes sense to (blindly) approve this MR.

I do wonder, though, what the impact of #309 will be in practice, on real data? There, exceptions may have escaped unnoticed as well.

darafferty:
I tested the changes on a few real images that I think are representative and found that they worked fine. If things are worse in some way due to the fixes, we can always adjust things ( e.g., the way the flagging is done) to make the results more similar to those before the fixes.

PR #315, opened 2026-06-16
Remove unused `np.where` from `rmsimage.py`

AlexKurek:
Since `np.where` turns out to be so slow (https://github.com/lofar-astron/PyBDSF/pull/311), it would be good to remove it where it is not needed.
.gaul files are identical, but currently the test will fail due to https://github.com/lofar-astron/PyBDSF/pull/314.


gmloose:
The function `numpy.where()` is also used in several other places, a quick `grep` reveals. I didn't look if some of these calls can also be removed safely, but it might be worth a check. If so, I have no objection if you extend this MR to cover the removal of all unnecessary calls to `numpy.where()`.

AlexKurek:
At the other places `np.where` has an actuall effect and is needed. Except this one:
https://github.com/lofar-astron/PyBDSF/blob/master/bdsf/rmsimage.py#L961
but I think there may also be a bug in this block (https://github.com/lofar-astron/PyBDSF/pull/317), so I think its safer to leave it for another PR.

AlexKurek:
Generally, `np.where` is redundant for boolean masking, checking element existence, or counting matching elements. Instead, directly use boolean indexing (arr[mask]), np.any(), or np.count_nonzero(). This avoids unnecessary memory overhead and improves speed.

gmloose (approved):

PR #316, opened 2026-06-17
Convert loaded image to native byteorder

bnikolic:
Makes a small (but repeatable) performance improvement of around 10% on intel/amd architecture for the SKA ref case I have and also allows use libraries which do not handle non-native byteorder. (NB Fits are always big-endian)

### DIFF
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -1073,6 +1073,17 @@ def get_kwargs(kwargs, key, typ, default):
 
     return obj
 
+def mknative(a):
+    """
+    Convert array a to native byteorder and return
+    """
+    if a.dtype.isnative:
+        return a
+    else:
+        # Strip the byte order from dtype string
+        nobodtype = str(a.dtype)[1:]
+        return a.astype("="+nobodtype)
+
 def read_image_from_file(filename, img, indir, quiet=False):
     """ Reads data and header from indir/filename.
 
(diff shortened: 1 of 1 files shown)


mhardcastle:
Just to note that we removed reordering in https://github.com/lofar-astron/PyBDSF/commit/a4a1bf9e2ac36197958a7823d423b49a449f4d44 because the overhead was making performance worse. So it's important to check that this different approach doesn't have the same problem...

bnikolic:
Thanks for pointing this out. 

Do you still have  the image / and/or testing script that showed the previous behaviour that you can share?  I can investigate 

The approach looks very similar but various things can have changed (such as performance of libraries with non-native orders) or perhaps the performance impact depends on the details of the use case...

mhardcastle:
I think it was just running on 20000 x 20000 LOFAR images (for single pointings, so we needed to load in two images, PB corrected and apparent flux). So not dissimilar to what I would imagine your use case is. A lot has changed in the code base since then as well as the libraries so I'd be willing to believe it might no longer be a problem.

bnikolic:
Yes, I'm using a 24k x 24k image for these tests, I captured it and the processing parameters from a SKA/LOFAR pipeline intermediate stage last year. 

I'm working on some other performance improvements so I propose we leave this in draft while doing all that just to see if something else related to this might come up

PR #317, opened 2026-06-18
Fix computing unmasked pixels in `rmsimage.py`

AlexKurek:
There is a mismatch between local slice coordinates and the global array.
The code uses `np.where(mask[a:b, c:d] == False)` to find unmasked pixels. This returns indices relative to the sliced box starting from (0,0).
A few lines down, these local coordinates are applied directly to the full image array: `arr[pix_unmasked]`.
It ends up calculating the mean and std for the top-left corner of the entire image instead of the intended sliced region (a:b, c:d).

In case of my 1.5 GB FITS the .gaul files are identical after this, so maybe it is a rare situation.

### DIFF
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -958,16 +958,19 @@ class Op_rmsimage(Op):
             if cnt > 198: cm = m; cr = r
             mean_map[i, j], rms_map[i, j] = cm, cr
         else:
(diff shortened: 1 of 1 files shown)


AlexKurek:
Needs modification after https://github.com/lofar-astron/PyBDSF/pull/335#discussion_r3531150519.

PR #319, opened 2026-06-18
Fix island pixel counting in `rmsimage.py` validation check

AlexKurek:
**Bug**
`N.where()` returns a tuple of coordinate arrays (one for each dimension). When passed to `N.size()`, we get (dimensions * pixel_count), so e.g., double of the actual pixel count for a 2D image.

**Fix**
Ditch `N.where()` to keep aper_mask as a boolean mask, and use .sum() to get the true pixel count.

This also may be rare, since for my 1.5 GB test image the .gaul files are identical.

### DIFF
--- a/bdsf/islands.py
+++ b/bdsf/islands.py
@@ -234,10 +234,10 @@ class Op_islands(Op):
             mean = img.mean_arr
             labels = func.make_src_mask(image.shape, isl_posn_pix, isl_radius_pix)
             if img.masked:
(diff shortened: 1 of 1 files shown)


darafferty (approved):
Nice catch! I think this code is only executed when you supply a list of islands by hand, so the bug was indeed likely rarely encountered. Anyway, thanks for the fix — the PR looks good.

PR #320, opened 2026-06-19
Fix mean value computation in `rmsimage.py`

AlexKurek:
If `goodcutout` has NaN-s, np.nansum will reject them, but len() will include them. Consequently, the computed mean value will be too low.

for my 1.5 GB FITS test image the .gaul files are identical after this.


gmloose (approved):
LGTM.

PR #321, opened 2026-06-21
Fix `shapelet_getroot` logic

AlexKurek:
The way it is now it will crash at:
`if dnew <= dold:`
since those variables are not defined - they were defined only at the previous `if`.


gmloose:
LGTM. @darafferty?

darafferty:
Yep, looks good to me as well.

darafferty (approved):

PR #322, opened 2026-06-21
Fix `try` block in `shape_findcen`

AlexKurek:
These variables are array indices, so they must by integers. But in Python 3 `/` returns a float, so it needs to be changed to `//` (integer division).


gmloose (approved):
LGTM. Thanks.

PR #323, opened 2026-06-21
Remove `fftconvolve`

AlexKurek:
`N.complex` was removed in Numpy 1.24.
Currently `fftconvolve` is not used since `scipy.signal.fftconvolve` is used, but still maybe better to fix it.
An alternative / next step could be to remove `fftconvolve` function.

### DIFF
--- a/bdsf/opts.py
+++ b/bdsf/opts.py
@@ -703,12 +703,6 @@ class Opts(object):
                                  "fitting is done on only the wavelet scale under "\
                                  "consideration.",
                              group = "atrous_do")
-    use_scipy_fft = Bool(True,
-                             doc = "Use fast SciPy FFT for convolution\n"\
-                                 "If True, the SciPy FFT function will be used instead "\
-                                 "of the custom version. The SciPy version is much "\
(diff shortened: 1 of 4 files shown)


gmloose (approved):
LGTM.

gmloose:
@darafferty, what about this unused `fftconvolve()` method? Should we remove it. The fix @AlexKurek made was trivial, so we could keep it. But if it is not used ...

bnikolic:
I've been looking at the convolution (https://github.com/lofar-astron/PyBDSF/pull/326 ) and it seems scipy is the simplest choice as well as having performance  which is competitive. The memory saving of inplace multiply  in the hand-rolled fftconvolve surprisingly has a performance impact (did not trace exactly why) so I think would be of interest in only quite limited circumstances.      So I would suggest removing the fftconvolve() is a good move.

darafferty:
Yes, I agree it is best to remove it, as I can't see it ever being used.

AlexKurek:
I have changed this PR to remove `fftconvolve`.

PR #325, opened 2026-06-22
Fix Numpy 2.5 incompatibility

AlexKurek:
`np.cross()` no longer supports 2D vectors:
https://github.com/numpy/numpy/releases/tag/v2.5.0

> numpy.cross no longer supports 2-dimensional vectors.

### DIFF
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -1577,7 +1577,9 @@ def area_polygon(points):
     area = 0.0
     for i in range(n_tri):
         p1, p2, p3 = N.array([cenx, ceny]), N.array([x[i], y[i]]), N.array([x[i+1], y[i+1]])
-        t_area= N.linalg.norm(N.cross((p2 - p1), (p3 - p1)))/2.
+        v1 = p2 - p1
+        v2 = p3 - p1
+        t_area = abs(v1[0] * v2[1] - v1[1] * v2[0]) / 2.0
         area += t_area
 
     return area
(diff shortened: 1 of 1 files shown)


AlexKurek:
Test script:
[compar.py](https://github.com/user-attachments/files/29732707/compar.py)

In Numpy 2.2.6:

> /home/akurek/compar.py:27: DeprecationWarning: Arrays of 2-dimensional vectors are deprecated. Use arrays of 3-dimensional vectors instead. (deprecated in NumPy 2.0)
>   t_area = N.linalg.norm(N.cross((p2 - p1), (p3 - p1)))/2.
> Area calculated by the old function: 4.0
> Area calculated by the new function: 4.0

darafferty:
Ah, I missed this PR (there are so many at the moment :) ). I will merge this one instead of mine, as it looks fine (and was first!).

AlexKurek:
Is it expected that there was no deprecation warning and that recent PR-s didn't crashed at build CI?

darafferty:
No, I'm not sure why it wasn't caught by the CI (I also didn't notice any deprecation warnings). Perhaps @gmloose has some ideas?

AlexKurek:
I think `area_polygon` is just not triggered in the test run and `DeprecationWarning` in the user code is ignored unless `PYTHONWARNINGS=default`. That would explain the situation.

AlexKurek:
...yes, I confirm. Doing:
`export PYTHONWARNINGS=default`
gives:

> /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/functions.py:1314: DeprecationWarning: Setting the shape on a NumPy array has been deprecated in NumPy 2.5.
> As an alternative, you can create a new view using np.reshape (with copy=False if needed).

PR #326, opened 2026-06-22
Turn on multi-threading for the wavelet convolve

bnikolic:
This enables multithreading already built in in scipy's pocketfft/ducc. It improves the performance of the wavelet calculation by about factor of 4 and overall performance for my SKA reference run by about 14%. 

I've fixed workers at 8 here since that seems to saturate the memory bus on testing CPUs yet isnt large number. Could be made variable.

### DIFF
--- a/bdsf/wavelet_atrous.py
+++ b/bdsf/wavelet_atrous.py
@@ -323,7 +323,8 @@ class Op_wavelet_atrous(Op):
         kern = N.outer(ff, ff)
         unmasked = N.nan_to_num(image)
         if use_scipy_fft:
-            im_new = scipy.signal.fftconvolve(unmasked, kern, mode='same')
+            with scipy.fft.set_workers(8):
+                im_new = scipy.signal.fftconvolve(unmasked, kern, mode='same')
         else:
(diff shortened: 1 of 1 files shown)


gmloose:
> I've fixed workers at 8 here since that seems to saturate the memory bus on testing CPUs yet isnt large number. Could be made variable.

I think I would prefer something like:
```
nworkers = min(8, nproc())
```
where `nproc()` should give you the number of processors available. Unfortunately, there's no such function (yet) in `PyBDSF`, but it could easily be added to, e.g., `bdsf/multi_proc.py`. That file already contains code to determine the number of available CPUs. So, it's just a matter of converting that code snippet into a function.

And yes, it would even be fancier if the hard-coded value of `8` were variable.

bnikolic:
I've adjusted this to use the numcores parameter already passed into the function, at least that would make it consistent with previous non-scipy function usage.   Could have a look at the `min` approach but may be a little while before I get a chance.

gmloose:
I would like to know when using multiple threads actually start to pay off. Unconditionally spinning up multiple threads might make the code slower instead of faster, unless arrays are large enough to benefit.

bnikolic:
OK. I need to check if there is already logic in scipy or ducc fft to adapt the number of workers to lower than the default if appropriate.

PR #327, opened 2026-06-26
Simplify adaptive RMS

AlexKurek:
This is one of the most costly parts of the PyBDSF process. A good place to start optimising it would be to remove `rank` and `connectivity` out of the loop.


gmloose:
LGTM.

PR #328, opened 2026-06-27
Fine-tune Gaussian fitting

AlexKurek:
Originally, if the algorithm failed to converge after 5 iterations, it resorted to fallback methods. The final fallback, moment analysis (`func.momanalmask_gaus`), simply fits a single, broad Gaussian to the remaining emission, which rarely gets flagged.  By increasing the limit to 15 iterations, we prevent this oversimplification. The algorithm now models complex islands with multiple Gaussians. As a byproduct, a few components might get flagged.

The 5 iteration limit is 15y old, now the computers are much faster.

### DIFF
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -543,7 +543,9 @@ def trans_gaul(q):
 def momanalmask_gaus(subim, mask, isrc, bmar_p, allpara=True):
     """ Compute 2d gaussian parameters from moment analysis, for an island with
         multiple gaussians. Compute only for gaussian with index (mask value) isrc.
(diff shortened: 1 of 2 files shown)


gmloose:
I'll let @darafferty decide. I guess the reference outputs need to be updated.

darafferty:
I think this is a great idea! As to the reference files, the tests pass for me now due to the higher tolerance — @AlexKurek can you pull in the recent changes? I think then the CI should pass.

AlexKurek:
Done and it passed the CI.

darafferty (approved):
Looks good!

PR #329, opened 2026-06-27
Remove unnecessary `np.where` from `gausfit.py`

AlexKurek:
Here:
`shedmask = N.where(watershed == ished+2, False, True) + isl.mask_active`
the program generates a mask by combining two conditions. The `N.where` construct used together with the + operator works (in NumPy, the + operator on boolean values behaves like a logical “OR”), but it is not optimal.


gmloose (approved):
LGTM.

PR #331, opened 2026-06-28
Fix external mask in `list_and_sort_gaussians`

AlexKurek:
1) `g.centre_pix` is a float, so it must be rounded before used for indexing.
2) From `make_casa_str` we see that the order in `g.centre_pix` is (x, y). But in Numpy the arrays are (row, column), so this order must be inverted.

### DIFF
--- a/bdsf/output.py
+++ b/bdsf/output.py
@@ -935,7 +935,9 @@ def list_and_sort_gaussians(img, patch=None, root=None,
                 gausflux = []
                 gausindx = []
             if use_mask:
-                patchnums.append(mask_labels[g.centre_pix[0], g.centre_pix[1]])
+                x_pix = int(round(g.centre_pix[0]))
+                y_pix = int(round(g.centre_pix[1]))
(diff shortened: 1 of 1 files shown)


gmloose:
I wonder why this doesn't break any tests. Swapping `x` and `y` should have an effect, I would say. @darafferty, do you have any idea.

AlexKurek:
I think this will change the list output files and the test compares only the images, am I right?

darafferty:
Maybe I'm missing something, but I think the mask image is reordered when it's read in with `func.read_image_from_file()` to have the order (x, y)?

Also, I guess the CI test doesn't cover this functionality, as no mask file has been specified (with the `bbs_patches_mask` option). But indeed the changes wouldn't affect the images anyway I guess.

AlexKurek:
Yes, I didn't noticed this, Im sorry. I have reverted the order change.

darafferty:
Great, thanks! Looks good to me.

darafferty (approved):

PR #332, opened 2026-06-28
Fix external mask in `list_and_sort_gaussians` (2)

AlexKurek:
**Bug**
Because `patchnums` is a standard Python list and p is an integer, `patchnums == p` does not check individual elements. Instead, it instantly evaluates to a single boolean False.

**Consequences**
NumPy forbid evaluating 0-dimensional scalars. Passing False to N.where() crashes the script with `ValueError: Calling nonzero on 0d arrays is not allowed`.

**Fix**
Cast the list to a NumPy array to force an element-by-element evaluation


gmloose (approved):
Good catch.

PR #333, opened 2026-06-30
Remove 14 unused functions from `functions.py`

AlexKurek:
All these functions are unused.

### DIFF
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -615,33 +615,6 @@ def get_errors(img, p, stdav, bm_pix=None, fixed_to_beam=False):
     return errors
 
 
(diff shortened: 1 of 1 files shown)


gmloose (commented):
I can confirm that these functions are not used by `PyBDSF` itself. However, these functions _are_ part of the public interface. It's hard to tell if anyone else is using these functions. What do you think, @darafferty? Shall we just remove them, or do you want to deprecate them first? The latter requires more effort from us, so it really depends how "friendly" we want to be to our users.

darafferty:
Hmm, good point. I doubt though that anyone is using these, so I would say we can remove them and, if someone objects, add them back as needed.

gmloose (approved):
OK, let's just remove them and see if people start to complain.

PR #334, opened 2026-06-30
Make `mapcoord_threaded` faster

AlexKurek:
**Chunking to reduce thread overhead**
**Before**: The function submitted a separate thread task for every single coordinate element in `axs[0]`.
**After**: The coordinates are split into `ncores` larger blocks using `np.array_split(axs[0], ncores)`. This reduces scheduling overhead and context switching.

**No threading for single core**
When `ncores` <= 1, skip ThreadPoolExecutor.

**Vectorized C operations**
Running on larger multi-coordinate chunks allows SciPy to perform vector operations.

A non-2D fallback is necessary since the optimised version only works for 2D.

For a (19845, 19845) image, Xeon 6238R (28 cores):

Before:
```
       28    0.001    0.000  244.910    8.747 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/rmsimage.py:26(mapcoord_threaded)
```
After:
```
       28    6.367    0.227   47.940    1.712 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/rmsimage.py:26(mapcoord_threaded)
```
5.1x faster.
The entire run is faster by ~15% if wavelets are on.
.gaul files are identical.

### DIFF
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -66,7 +66,7 @@ def mapcoord_threaded(a, axs, *args, ncores=None, **kwargs):
 
 
     if ncores is None:
-        ncores = min(32, (mp.nproc or 1) + 4)
+        ncores = mp.nproc()
 
     output = kwargs.pop("output", None)
     order = kwargs.get("order", 3)
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -65,8 +65,9 @@ def mapcoord_threaded(a, axs, *args, ncores=None, **kwargs):
         )
 
 
+    available_cpus = len(os.sched_getaffinity(0))
     if ncores is None:
-        ncores = min(32, (os.cpu_count() or 1) + 4)
+        ncores = min(32, (available_cpus or 1) + 4)
 
     output = kwargs.pop("output", None)
     order = kwargs.get("order", 3)
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -76,8 +76,8 @@ def mapcoord_threaded(a, axs, *args, ncores=8, **kwargs):
     if order > 1:
         a = ndimage.spline_filter(a, order, output=np.float64, mode=mode)
 
-    # Fallback for non-2D cases
-    if len(axs) != 2:
+    # Fallback for 3D cases
+    if len(axs) > 2:
         if output is None:
             shape = [len(axs[1]), len(axs[0])] + [len(ax) for ax in axs[2:]]
             out_dtype = np.float64 if order > 1 else a.dtype
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -80,11 +80,11 @@ def mapcoord_threaded(a, axs, *args, ncores=8, **kwargs):
     # Limit the tile size: 
     # max 1000: to avoid evicting data from the CPU L3 cache (~16 MB)
     # min 100:  so the Python array creation overhead doesn't exceed computation time
-    TILE_SIZE = max(100, min(1000, ideal_tile_size))
+    tile_size = max(100, min(1000, ideal_tile_size))
     
     # Generate coordinate ranges for individual tiles
-    ranges_0 = [(i, min(i + TILE_SIZE, len0)) for i in range(0, len0, TILE_SIZE)]
-    ranges_1 = [(i, min(i + TILE_SIZE, len1)) for i in range(0, len1, TILE_SIZE)]
+    ranges_0 = [(i, min(i + tile_size, len0)) for i in range(0, len0, tile_size)]
+    ranges_1 = [(i, min(i + tile_size, len1)) for i in range(0, len1, tile_size)]
     
     # Prepare coordinate pairs to eliminate task assignment delays
     tasks = [(r0, r1) for r1 in ranges_1 for r0 in ranges_0]
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -67,7 +67,7 @@ def mapcoord_threaded(a, axs, *args, ncores=None, **kwargs):
 
     available_cpus = len(os.sched_getaffinity(0))
     if ncores is None:
-        ncores = min(32, (available_cpus or 1) + 4)
+        ncores = min(32, (mp.nproc or 1) + 4)
 
     output = kwargs.pop("output", None)
     order = kwargs.get("order", 3)
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -65,7 +65,6 @@ def mapcoord_threaded(a, axs, *args, ncores=None, **kwargs):
         )
 
 
-    available_cpus = len(os.sched_getaffinity(0))
     if ncores is None:
         ncores = min(32, (mp.nproc or 1) + 4)
 
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -35,10 +35,10 @@ def mapcoord_threaded(a, axs, *args, ncores=8, **kwargs):
     if ncores is None:
         ncores = min(32, (os.cpu_count() or 1) + 4)
 
-    output = kwargs.get("output", None)
-    kwargs["output"] = None
-    # Prefilter only once to awoid repeated work in the workers. See
-    # _interpolation.py in scipy
+    # Retrieve output and remove it from kwargs to manage it manually
+    output = kwargs.pop("output", None)
+    
+    # Pre-filtering performed once
     order = kwargs.get("order", 3)
     if order > 1:
         a = ndimage.spline_filter(a, order,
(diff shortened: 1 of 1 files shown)


AlexKurek:
https://github.com/lofar-astron/PyBDSF/pull/334/commits/988673632f8f436a305761bcc94b4bea472835ee introduces pre-allocation of memory for the whole result instead of joining chunks.
```
       28    0.002    0.000   30.571    1.092 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/rmsimage.py:26(mapcoord_threaded)
```
Now its 8x faster.

AlexKurek:
Now its 11x faster:
```
       28    0.003    0.000   22.427    0.801 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/rmsimage.py:26(mapcoord_threaded)
```

bdsf/rmsimage.py
```diff
@@ -32,36 +32,99 @@ def mapcoord_threaded(a, axs, *args, ncores=8, **kwargs):
     to equivalent of meshgrid(*axs)
 
     """
-    output = kwargs.get("output", None)
-    kwargs["output"] = None
-    # Prefilter only once to awoid repeated work in the workers. See
-    # _interpolation.py in scipy
+    if ncores is None:
+        ncores = min(32, (os.cpu_count() or 1) + 4)
```
gmloose: The function `os.cpu_count()` will always return the number of CPUs present in the machine. It's better to ask how many CPUs are available for the current process. There is some logic in `bdsf/multi_proc.py`. Maybe we should turn that into a function `nproc()`, and use that instead.

bdsf/rmsimage.py
```diff
@@ -32,36 +32,99 @@ def mapcoord_threaded(a, axs, *args, ncores=8, **kwargs):
     to equivalent of meshgrid(*axs)
 
     """
-    output = kwargs.get("output", None)
-    kwargs["output"] = None
-    # Prefilter only once to awoid repeated work in the workers. See
-    # _interpolation.py in scipy
+    if ncores is None:
```
gmloose, line 68: Hm, `ncores` is also passed as argument, which defaults to `8`. It's unlikely that it will ever be `None`, unless someone purposely wants to wreak havoc. If you want to be "clever" when the user does not supply `ncores` as argument, it makes more sense to set in to `None` by default in line 26.

bdsf/rmsimage.py
```diff
@@ -32,36 +32,99 @@ def mapcoord_threaded(a, axs, *args, ncores=8, **kwargs):
...
+        if output is None:
+            shape = [len(axs[1]), len(axs[0])] + [len(ax) for ax in axs[2:]]
+            out_dtype = np.float64 if order > 1 else a.dtype
+            output = np.empty(shape, dtype=out_dtype)
+        
+        chunks = np.array_split(axs[0], ncores)
+        chunk_sizes = [len(c) for c in chunks]
+        indices = [0] + np.cumsum(chunk_sizes).tolist()
+
+        def tworker_fallback(i, cl1):
```
gmloose: I would prefer to have nested functions defined at the top of this function.

bdsf/rmsimage.py
```diff
@@ -32,36 +32,99 @@ def mapcoord_threaded(a, axs, *args, ncores=8, **kwargs):
...
+    TILE_SIZE = max(100, min(1000, ideal_tile_size))
+    
+    # Generate coordinate ranges for individual tiles
+    ranges_0 = [(i, min(i + TILE_SIZE, len0)) for i in range(0, len0, TILE_SIZE)]
+    ranges_1 = [(i, min(i + TILE_SIZE, len1)) for i in range(0, len1, TILE_SIZE)]
+    
+    # Prepare coordinate pairs to eliminate task assignment delays
+    tasks = [(r0, r1) for r1 in ranges_1 for r0 in ranges_0]
+
+    def tworker_tiled(r0, r1):
```
gmloose: I would prefer to have nested functions defined at the top of this function.

bdsf/rmsimage.py
```diff
@@ -32,36 +32,99 @@ def mapcoord_threaded(a, axs, *args, ncores=8, **kwargs):
...
-        if output is not None:
-            np.copyto(output, res)
-        return res
-
+        a = ndimage.spline_filter(a, order, output=np.float64, mode=mode)
+
+    # Fallback for non-2D cases
+    if len(axs) != 2:
+        if output is None:
+            shape = [len(axs[1]), len(axs[0])] + [len(ax) for ax in axs[2:]]
```
gmloose, line 82: This will fail if `len(axs)` is `0` or `1`. Maybe the check in line 47 should be changed to `if len(axs) > 2:`?

bdsf/rmsimage.py
```diff
@@ -32,36 +32,99 @@ def mapcoord_threaded(a, axs, *args, ncores=8, **kwargs):
     to equivalent of meshgrid(*axs)
 
     """
```
gmloose, line 34: Should we maybe have a pre-condition check of `axs`? This function makes assumptions about its dimensions, which are nowhere guaranteed.

bdsf/rmsimage.py
```diff
@@ -32,36 +32,99 @@ def mapcoord_threaded(a, axs, *args, ncores=8, **kwargs):
...
+    if output is None:
+        output = np.empty((len1, len0), dtype=np.float32)
+
+    # Aim to split the image to generate enough tasks for all cores
+    ideal_tile_size = int(max(len0, len1) / (np.sqrt(ncores) or 1))
+    
+    # Limit the tile size: 
+    # max 1000: to avoid evicting data from the CPU L3 cache (~16 MB)
+    # min 100:  so the Python array creation overhead doesn't exceed computation time
+    TILE_SIZE = max(100, min(1000, ideal_tile_size))
```
gmloose: Just a nit-pick. But why are you using an UPPERCASED variable name. This is usually reserved for (global) constants.

gmloose (commented):
I didn't see any obvious mistakes, so I guess this is okay. Could you have a look too, @darafferty ?

bdsf/rmsimage.py
AlexKurek: There is no such check in the original version. It is not user-supplied, so I think its safe.

bdsf/rmsimage.py
AlexKurek: Good catch, thank you, there is a difference. From what I can find, it can be done this way:
```
import os
available_cpus = len(os.sched_getaffinity(0))
```
From Python 3.13 there is also:
`available_cpus = os.process_cpu_count()`
I have added `available_cpus` in https://github.com/lofar-astron/PyBDSF/pull/334/commits/95b7430477bc9c434cb112ca9ac9aa65dc7d9b9e.

darafferty:
It looks good to me. This has always been one of the slower bits (especially as the images get larger), so speed-ups here are very welcome.

bdsf/rmsimage.py
```diff
@@ -32,8 +32,42 @@ def mapcoord_threaded(a, axs, *args, ncores=8, **kwargs):
...
+        
+        ndimage.map_coordinates(
+            a, coords, 
+            prefilter=False, 
+            output=output[out_slice], 
+            *args, **kwargs
+        )
+
+
+    available_cpus = len(os.sched_getaffinity(0))
```
gmloose: This doesn't work on MacOS. I propose we upgrade the logic in `bdsf/multi_proc.py` to a function that can be imported and used. Something like:
```
def nproc():
    """
    Return the number of CPU cores _available_ to the current process, similar
    to what the Linux `nproc` command does. This can be less than the total
    number of CPU cores in the machine.
    NOTE: This function uses `os.sched_getaffinity()`, which is not available
    on every OS. Use `multiprocessing.cpu_count()` as fall-back.
    """
    try:
        return len(os.sched_getaffinity(0))
    except AttributeError:
        return multiprocessing.cpu_count() or 1
```

AlexKurek: This should work on Mac:

```
cpu_count = psutil.cpu_count(logical=False) # cores
cpu_count_logical = psutil.cpu_count(logical=True) # threads

available = len(psutil.Process().cpu_affinity())
```

gmloose: Hm, according to the docs, this is only available on Linux, Windows, and FreeBSD. But I have another objection against using `psutil`: it's not a standard library, and I'd prefer not to pull in a dependency on `pustil` only for this.

AlexKurek: So maybe consider adding this Python 3.13 thing:
```
available_cpus = os.process_cpu_count()
```
to the try block in your new function.

gmloose: I could, but I looked at the implementation, and it basically just does what my function does. Here it is:
```
if _exists('sched_getaffinity') and sys._get_cpu_count_config() < 0:
    def process_cpu_count():
        """
        Get the number of CPUs of the current process.

        Return the number of logical CPUs usable by the calling thread of the
        current process. Return None if indeterminable.
        """
        return len(sched_getaffinity(0))
else:
    # Just an alias to cpu_count() (same docstring)
    process_cpu_count = cpu_count
```

bdsf/rmsimage.py
```diff
@@ -32,36 +32,100 @@ def mapcoord_threaded(a, axs, *args, ncores=8, **kwargs):
...
+        ndimage.map_coordinates(
+            a, coords, 
+            prefilter=False, 
+            output=output[out_slice], 
+            *args, **kwargs
+        )
+
+
+    if ncores is None:
+        ncores = min(32, (mp.nproc or 1) + 4)
```
gmloose: Meanwhile, I've added a method `nproc()` to `bdsf/multi_proc.py`. So, I'd suggest something like the following:
```suggestion
        ncores = mp.nproc()
```
Feel free to make any adjustments if you want to use more of fewer cores than available.

AlexKurek: This my recent limit to 32 was arbitrary. Actually, maybe lets not impose harcoded limits? It should always be faster than the original.

darafferty (approved):
Just gave this a final check, and it looks good. Nice work!

PR #335, opened 2026-07-04
More robust statistics of a very small RMS window

AlexKurek:
If a window contains very few unmasked pixels (between 6 and 20), the code abandons sigma clipping in favor of mean and standard deviation. This reduces robustness against sources in crowded regions (like probably the reference image in CI).

Solution: robust statistics that are less sensitive to outliers: median and Median Absolute Deviation (MAD). For distributions that are approximately Gaussian, the standard deviation can be estimated by multiplying MAD by 1.4826.

It prevents a single defective or contaminated pixel from distorting the RMS and mean map.

For my 20x x 20k test image the .gaul files are identical.

### DIFF
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -965,9 +965,18 @@ class Op_rmsimage(Op):
                 if cnt > 198: cm = m; cr = r
                 mean_map[i, j], rms_map[i, j] = cm, cr
             else:
-                if npix_unmasked > 5: # just find simple mean/rms
-                    cm = np.mean(arr[pix_unmasked])
-                    cr = np.std(arr[pix_unmasked])
+                if npix_unmasked > 5: # use robust median and MAD instead of mean/std
+                    valid_pixels = arr_slice[unmasked_mask]
+                    cm = np.median(valid_pixels)
+                    
+                    # Calculate standard deviation estimated from MAD
+                    # MAD = median(|x - median(x)|). The scale factor for a Gaussian distribution is 1.4826
+                    mad = np.median(np.abs(valid_pixels - cm))
+                    cr = 1.4826 * mad
+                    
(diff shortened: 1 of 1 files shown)


bdsf/rmsimage.py
```diff
@@ -965,9 +965,18 @@ def for_masked(self, mean_map, rms_map, mask, arr, ind, kappa, co):
                 if cnt > 198: cm = m; cr = r
                 mean_map[i, j], rms_map[i, j] = cm, cr
             else:
-                if npix_unmasked > 5: # just find simple mean/rms
-                    cm = np.mean(arr[pix_unmasked])
-                    cr = np.std(arr[pix_unmasked])
+                if npix_unmasked > 5: # use robust median and MAD instead of mean/std
+                    valid_pixels = arr_slice[unmasked_mask]
```
gmloose: Maybe I missed something, but where are `arr_slice` and `unmasked_mask` defined?

bdsf/rmsimage.py
```diff
@@ -965,9 +965,18 @@ def for_masked(self, mean_map, rms_map, mask, arr, ind, kappa, co):
...
+                    cm = np.median(valid_pixels)
+                    
+                    # Calculate standard deviation estimated from MAD
```
gmloose:
```suggestion
                    # Calculate standard deviation estimated from Median Absolute Deviation (MAD)
```

bdsf/rmsimage.py
AlexKurek: Im sorry, locally I merged my other PR-s for this module and this is where those variables are defined. Thank you for cathing this. I will fix it soon.

AlexKurek: Fixed. Now it also partially solves https://github.com/lofar-astron/PyBDSF/pull/317.

gmloose (approved):
LGTM.

PR #337, opened 2026-07-07
Simplify `for_masked`

AlexKurek:
Make `for_masked` just call `for_masked_mp` for calculations, as all of them were repeated. Also took the comments to `for_masked_mp`.


gmloose (approved):
LGTM. Thanks for another simplification.

PR #338, opened 2026-07-07
Remove backport to `shutil_get_terminal_size`

AlexKurek:
This is not needed since Python 3.3:
https://pypi.org/project/backports.shutil_get_terminal_size/
It may event be shadowing the native `shutil_get_terminal_size`.


gmloose:
LGTM.

PR #339, opened 2026-07-07
Fix extended source filtering in adaptive rms box calculation

AlexKurek:
Later the code assumes that the indices in these lists correspond to each other:

```
bright_indx = isl_maxposn.index(isl_maxposn_lowthresh)

isl_area_highthresh[bright_indx]
isl_size_bright[bright_indx]
```

Presently `isl_area_highthresh` is not cleaned at each iteration, so the computed island areas do not correspond to the correct islands.


gmloose (commented):
I think you're right, but I would like a confirmation from @darafferty.

darafferty:
Yes, indeed, it was incorrect. Thanks for the fix!

PR #340, opened 2026-07-08
Fix call to numpy.cross() function

darafferty:
This PR updates the calls to the `numpy.cross()` function to be compatible with numpy v2.5.0 and later.


AlexKurek:
This is related: https://github.com/lofar-astron/PyBDSF/pull/325.

darafferty:
Above PR merged instead, so closing this one.

PR #341, opened 2026-07-08
Fix Numpy 2.5 deprecation

AlexKurek:
Fix for:
> /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/functions.py:1314: DeprecationWarning: Setting the shape on a NumPy array has been deprecated in NumPy 2.5.
> As an alternative, you can create a new view using np.reshape (with copy=False if needed).
>   data.shape = data.shape[0:4] # trim unused dimensions (if any)

Somewhat related: https://github.com/lofar-astron/PyBDSF/pull/325#issuecomment-4917658409.


gmloose (approved):
LGTM.

PR #342, opened 2026-07-09
Fix ResourceWarning

AlexKurek:
This warning:

> <sys>:0: ResourceWarning: unclosed file <_io.BufferedReader name='image_full_ampphase_di_m.NS_shift.int.facetRestored.fits'>

shows up if `export PYTHONWARNINGS=default`.
It occurs because the file is not closed if any error or exception occurs.
To ensure that the file is always closed, move the file opening and processing into `try...finally`.

BTW. There is also this, but I dont know how to fix it.

> --> Using adaptive scaling of rms_box
> /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/multiprocessing/popen_fork.py:76: DeprecationWarning: This process (pid=2622643) is multi-threaded, use of fork() may lead to deadlocks in the child.
>   self.pid = os.fork()

### DIFF
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -1109,214 +1109,217 @@ def read_image_from_file(filename, img, indir, quiet=False):
         img._reason = f'File {image_file} does not exist'
         return None
 
-    # If img.use_io is set, then use appropriate io module
-    if img.use_io:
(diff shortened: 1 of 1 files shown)


AlexKurek:
I'm sorry, but this doesn't remove the warning. I thought it did. I'm closing it since a different fix is needed.

PR #344, opened 2026-07-12
Remove dead code from `treshold.py`

AlexKurek:
This is unused.


darafferty:
Yep, good to remove these.

PR #345, opened 2026-07-12
Fix FDR algorithm

AlexKurek:
The code sorted the p-values in descending order, so it evaluated the wrong condition. This broke the FDR, turning it into a flawed hard threshold that allowed too many false positive pixels (noise).

Fix: Sort the p-values in ascending order and iterate backward (from the total number of pixels N down to 1) to find the largest index k that satisfies Benjamini-Yekutieli FDR condition.


gmloose (approved):
LGTM. I checked with @darafferty, and he agrees with what you're saying.

PR #346, opened 2026-07-12
Add warning to `treshold.py`

AlexKurek:
Add:

> mylogger.userinfo(mylog, "Detection threshold lies outside the calibrated source-count range.")
> mylogger.userinfo(mylog, "Source count estimate in threshold method selection is extrapolated.")

It would be good to see how often this happens in the LOFAR era (the used model is very old).
Eg. it happens for my test image.

### DIFF
--- a/bdsf/threshold.py
+++ b/bdsf/threshold.py
@@ -101,6 +101,8 @@ class Op_threshold(Op):
         for i,s in enumerate(scflux):
             if s < smin_L:
                 index = i
+                mylogger.userinfo(mylog, "Detection threshold lies outside the calibrated source-count range.")
+                mylogger.userinfo(mylog, "Source count estimate is extrapolated.")
                 break
         n1 = scnum[index]; n2 = scnum[-1]
         s1 = scflux[index]; s2 = scflux[-1]
--- a/bdsf/threshold.py
+++ b/bdsf/threshold.py
@@ -34,7 +34,7 @@ class Op_threshold(Op):
         sq2  = sqrt(2)
 
         if img.opts.thresh is None:
-            source_p = self.get_srcp(img)
+            source_p = self.get_srcp(img, mylog)
             cutoff = 5.0
             false_p = 0.5*erfc(cutoff/sq2)*size
             if false_p < opts.fdr_ratio*source_p:
@@ -84,7 +84,7 @@ class Op_threshold(Op):
         img.completed_Ops.append('threshold')
         return img
 
-    def get_srcp(self, img):
+    def get_srcp(self, img, mylog):
         from . import sourcecounts as sc
         fwsig = const.fwsig
         cutoff = 5.0
--- a/bdsf/threshold.py
+++ b/bdsf/threshold.py
@@ -102,7 +102,7 @@ class Op_threshold(Op):
             if s < smin_L:
                 index = i
                 mylogger.userinfo(mylog, "Detection threshold lies outside the calibrated source-count range.")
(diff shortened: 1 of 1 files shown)


bdsf/threshold.py
```diff
@@ -99,6 +99,8 @@ def get_srcp(self, img):
         for i,s in enumerate(scflux):
             if s < smin_L:
                 index = i
+                mylogger.warning(mylog, "Detection threshold lies outside the calibrated source-count range.")
+                mylogger.warning(mylog, "Source count estimate in threshold method selection is extrapolated.")
```
gmloose: I may be missing the point here. But why would `mylog.warning()` not work here?

gmloose (commented):
I'm not a big fan of passing `logger` objects around. Also, I don't completely understand the semantics here (see my comment).

bdsf/threshold.py
AlexKurek: I dont really know how logging works, but for
`mylog.warning("Detection threshold lies outside the calibrated source-count range.")`
I got:
```
  File "/home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/threshold.py", line 107, in get_srcp
    mylog.warning("Detection threshold lies outside the calibrated source-count range.")
    ^^^^^
NameError: name 'mylog' is not defined
```
Do you have a preference how this should be done?

gmloose: That's weird. You pass it as argument in line 90. 🤔 
I understand more or less how logging works, but I don't fully understand what the original authors had in mind when they wrote `mylogger.py`. I would really need to dive into that. Maybe @darafferty knows, but it could very well even predate his contributions. It definitely predates 2012 when I created PyBDSF as separate repository (SVN at that time).

AlexKurek: I mean this doesnt work after removing passing it as argument in line 90 ;) since you suggested not to pass it in the argument.
Maybe it will work if we do:
`mylog = mylogger.logging.getLogger("PyBDSF."+img.log+"Threshold ")`
in this function (maybe this line has to be done in each function sperately).


A bit of explanation:
https://github.com/lofar-astron/PyBDSF/blob/master/bdsf/sourcecounts.py is probably very old and refers to 1.4 GHz. LOFAR observes different galaxies, more AGNs, etc. So I wonder if this algorithm actually works these days. A good starting point would be to add this warning.

AlexKurek: cProfile shows that `get_srcp` is executed only once and it takes 0% time. So maybe it doesnt really matter that `mylog` is passed there.

gmloose: > I mean this doesnt work after removing passing it as argument in line 90 ;) since you suggested not to pass it in the argument. Maybe it will work if we do: `mylog = mylogger.logging.getLogger("PyBDSF."+img.log+"Threshold ")` in this function (maybe this line has to be done in each function sperately).

Sorry, I misunderstood that. One possible option would be to have an `__init__()` that defines `self.mylog` or `self.logger` (my preference). Then you can you use `self.logger` in the `__call__()` method. That is a bit cleaner than passing `mylog` around (IMHO).

AlexKurek: I have tested it the way it is now (https://github.com/lofar-astron/PyBDSF/pull/346/commits/311ba9f81e502cf80cc411e28a4bc0db6e3a4ab7) and its working. Is this how you see it? Looks cleaner indeed, just I had to modify a few more lines also.

gmloose (approved):

PR #348, opened 2026-07-14
Clean `interface.py`

AlexKurek:
- leftovers from Python
- in `for i, key in enumerate(kwargs):` `i` was not used
- `word.strip()` was not doing anything

### DIFF
--- a/bdsf/interface.py
+++ b/bdsf/interface.py
@@ -5,14 +5,7 @@ interactive environment such as IPython. Many are also used by the
 custom IPython shell defined in pybdsf.
 
 """
-from __future__ import print_function
-from __future__ import absolute_import
-
-try:
-    # For Python 2, use raw_input() for input()
-    input = raw_input
-except NameError:
-    pass
+import os
 
 
 def process(img, **kwargs):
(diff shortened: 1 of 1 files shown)


bdsf/interface.py
```diff
@@ -355,11 +346,7 @@ def load_pars(filename):
     Returns None (and original error) if no file can be loaded successfully.
     """
     from .image import Image
-    from . import mylogger
-    try:
-        import cPickle as pickle
-    except ImportError:
-        import pickle
+    import pickle
```
gmloose: I would move this import to the top of the file.

bdsf/interface.py
```diff
@@ -381,12 +368,7 @@ def save_pars(img, savefile=None, quiet=False):
 
     The save file is a "pickled" opts dictionary.
     """
-    try:
-        import cPickle as pickle
-    except ImportError:
-        import pickle
-    from . import tc
-    import sys
+    import pickle
```
gmloose: I would move this import to the top of the file.

gmloose (commented):
LGTM. But there are merge conflicts.

darafferty (approved):
Indeed, looks good — definitely an improvement!

PR #349, opened 2026-07-14
Fix formatting of `beam_spectrum` option in `print_opts()`

AlexKurek:
Reset the temporary list in round_list_of_tuples() to correctly format multiple tuples.


gmloose (approved):
LGTM.

PR #350, opened 2026-07-14
Fix: subtract background when computing `total_flux`

AlexKurek:
If the background is not subtracted, the calculated total flux of the island is usually slightly inaccurate.
Figure_1

### DIFF
--- a/bdsf/islands.py
+++ b/bdsf/islands.py
@@ -350,8 +350,11 @@ class Island(object):
         in_bbox_and_unmasked = N.where(~N.isnan(bbox_mean_im))
         self.mean = bbox_mean_im[in_bbox_and_unmasked].mean()
         self.islmean = bbox_mean_im[in_bbox_and_unmasked].mean()
-        valid_pixels = ~self.mask_active & ~N.isnan(self.image) # pixels that are both: inside the island and not NaN
-        self.total_flux = N.nansum(self.image[valid_pixels])/beamarea
+        # Create a mask for pixels that are:
+        # a) within the island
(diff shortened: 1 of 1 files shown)


bdsf/islands.py
```diff
@@ -350,8 +350,11 @@ def __init__(self, img, mask, mean, rms, labels, bbox, idx,
         in_bbox_and_unmasked = N.where(~N.isnan(bbox_mean_im))
         self.mean = bbox_mean_im[in_bbox_and_unmasked].mean()
         self.islmean = bbox_mean_im[in_bbox_and_unmasked].mean()
-        valid_pixels = ~self.mask_active & ~N.isnan(self.image) # pixels that are both: inside the island and not NaN
-        self.total_flux = N.nansum(self.image[valid_pixels])/beamarea
+        # Create a mask for pixels that are:
+        # a) within the island
+        # b) not NaN in both the image and the background map
+        valid_pixels = ~self.mask_active & ~N.isnan(self.image) & ~N.isnan(bbox_mean_im)
```
gmloose, line 363: Isn't `~N.isnan(bbox_meam_im)` the same as `in_bbox_and_unmasked`, set in line 350?

gmloose (commented):
I think this is OK. I'll let @darafferty have a look as well.

bdsf/islands.py
AlexKurek: I have modified this lines slightly.

darafferty:
Looks OK to me too.

gmloose (approved):
That's cleaner, and probably faster too. 😄

PR #352, opened 2026-07-15
Fix island IDs map when `coords_to_isl` is used

AlexKurek:
**Problem**
The code uses the addition operator to assign island values. This fails in areas where islands overlap, as their ID values are summed instead of being overwritten.

**When it occurs**
When 2 or more islands overlap when constructing islands using `coords_to_isl`.

**Fix**
Replaced the addition with `numpy.where`. This ensures that overlapping pixels are overwritten with the latest island ID instead of accumulating ID values.


gmloose (approved):
LGTM.

PR #354, opened 2026-07-16
Add nproc() method

gmloose:
Added `nproc()` method to determine the number of CPUs _available_ to the current process. If that cannot be determined, then return the number of CPUs present in the system.

### DIFF
--- a/bdsf/multi_proc.py
+++ b/bdsf/multi_proc.py
@@ -17,13 +17,21 @@ import traceback
 
 import numpy
 
-# Try to determine the number of CPU cores _available_ to the current process,
-# similar to what the Linux `nproc` command does. If that fails, return the
-# total number of CPU cores in the machine.
-try:
-    _ncpus = len(os.sched_getaffinity(0))
-except AttributeError:
-    _ncpus = multiprocessing.cpu_count()
+def nproc():
+    """
+    Return the number of CPU cores _available_ to the current process, similar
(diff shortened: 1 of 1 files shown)


bnikolic:
I looked into this one since you mention this function it in the other PR. It totally does not matter in this (or probably any other) case but it can done without using exception with something like:

```
if f := getattr(os, "sched_getaffinity", None): 
   return len (f(0))	
else:
   return os.cpu_count() or 1
```

(I guess many years ago there was a superstition if not reality that exceptions are expensive and so I got used to not using them except when the stack actually needs unwinding a few levels)

darafferty (approved):
Looks good to me.

gmloose:
> I looked into this one since you mention this function it in the other PR. It totally does not matter in this (or probably any other) case but it can done without using exception with something like:
> 
> ```
> if f := getattr(os, "sched_getaffinity", None): 
>    return len (f(0))	
> else:
>    return os.cpu_count() or 1
> ```
> 
> (I guess many years ago there was a superstition if not reality that exceptions are expensive and so I got used to not using them except when the stack actually needs unwinding a few levels)

Yep. That's another solution. Exceptions are Python's bread and butter. Even `for` loops use exceptions (`StopIteration`). I think the lore more applies to C++ where exceptions could have significant overhead, but I don't think it's so problematic anymore nowadays.

PR #355, opened 2026-07-16
Increase numeric tolerance

gmloose:
The numeric tolerance of 0.01 turned out to be slightly too strict. Setting it to 0.03 avoids a lot of false alarms.


darafferty (approved):
Looks good!

PR #357, opened 2026-07-17
Fix TOCTOU race in directory creation

### LINKED ISSUE #356: `FileExistsError` crash when running `process_image` from concurrent processes
Running `bdsf.process_image()` from multiple concurrent processes with output directories under a shared parent intermittently crashes one or more workers. We found this issue across SLURM job arrays sharing an output directory tree (unfortunately missing specific job logs, we patched this locally months ago), but it can theoretically also trigger under local multiprocessing pools too. The cause is the check-then-create pattern used for directory creation, e.g. `functions.py` in `set_up_output_paths()`:

```python
if not os.path.exists(output_basedir):
    os.makedirs(output_basedir)
```

Between the check and the `makedirs`, another worker creates the directory, and the loser crashes, killing its whole batch. The window is most likely to be hit at startup, when concurrent workers are synchronised and all pass through this code at once. On NFS (where cluster users typically run), `os.path.exists()` can also return stale results due to attribute caching, so the check is unreliable even beyond the race window. The same pattern occurs at ~17 sites across the package.

`islands.py` already uses the race-free form I am planning to implement (`os.makedirs(..., exist_ok=True)`, added in #209); the fix is to apply it consistently across the whole program. I have a patch ready and will open a PR shortly.

Siphlygon:
Fixes #356

Concurrent `process_image()` calls sharing an output directory tree (e.g. SLURM array tasks writing to a common parent) can crash with `FileExistsError`. Directory creation throughout the package uses a check-then-create pattern:

```python
if not os.path.exists(d):
    os.makedirs(d)
```

which has a race window between the check and the call - see #356. This PR converts the remaining sites to `os.makedirs(d, exist_ok=True)`, which is already used in `islands.py` (#209) and avoids the race condition entirely. All lines changed are shown in the table below.

| File | Lines |
|---|---|
| `bdsf/functions.py` | 1334, 1397, 1552, 2235 |
| `bdsf/cleanup.py` | 37 |
| `bdsf/make_residimage.py` | 71, 79 |
| `bdsf/output.py` | 24, 36, 831 |
| `bdsf/readimage.py` | 85, 141 |
| `bdsf/rmsimage.py` | 383, 391, 399 |
| `bdsf/wavelet_atrous.py` | 56, 58, 60 |

Every site was already guarded by an existence check, so sequential behaviour is unchanged, meaning only the concurrent crash is removed. `exist_ok=True` still raises if the path exists as a non-directory, so genuine misconfigurations surface exactly as before. I've dropped the now-redundant `if not exists` guards to match the `islands.py` style (one line instead of two); happy to keep them if a minimal diff is preferred.

A TOCTOU race isn't reproducible deterministically, so no regression test is included. The existing suite exercises every touched path sequentially and confirms no behaviour change there.

### DIFF
--- a/bdsf/cleanup.py
+++ b/bdsf/cleanup.py
@@ -34,7 +34,7 @@ class Op_cleanup(Op):
 
             from math import log10
             bdir = img.basedir + '/misc/'
-            if not os.path.isdir(bdir): os.makedirs(bdir, exist_ok=True)
+            os.makedirs(bdir, exist_ok=True)
             im_mean = img.clipped_mean
             im_rms = img.clipped_rms
             low = 1.1*abs(img.min_value)
--- a/bdsf/cleanup.py
+++ b/bdsf/cleanup.py
@@ -34,7 +34,7 @@ class Op_cleanup(Op):
 
             from math import log10
             bdir = img.basedir + '/misc/'
-            if not os.path.isdir(bdir): os.makedirs(bdir)
+            if not os.path.isdir(bdir): os.makedirs(bdir, exist_ok=True)
             im_mean = img.clipped_mean
             im_rms = img.clipped_rms
             low = 1.1*abs(img.min_value)
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -1390,7 +1390,7 @@ def write_image_to_file(use, filename, image, img, outdir=None,
         send_fits_image(img.samp_client, img.samp_key, 'PyBDSF image', tfile.name)
     else:
         # Write image to FITS file
-        if outdir is None:
+        if not outdir:
             outdir = img.indir
         os.makedirs(outdir, exist_ok=True)
         outfilename = os.path.join(outdir, filename)
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -1330,8 +1330,7 @@ def convert_casacore_header(casacore_image, tmpdir):
     except ImportError as err:
         import pyfits
 
-    if not os.path.exists(tmpdir):
-        os.makedirs(tmpdir)
+    os.makedirs(tmpdir, exist_ok=True)
     tfile = tempfile.NamedTemporaryFile(delete=False, dir=tmpdir)
     casacore_image.tofits(tfile.name)
     hdr = pyfits.getheader(tfile.name)
@@ -1393,8 +1392,7 @@ def write_image_to_file(use, filename, image, img, outdir=None,
         # Write image to FITS file
         if outdir is None:
             outdir = img.indir
-        if not os.path.exists(outdir) and outdir != '':
-            os.makedirs(outdir)
+        os.makedirs(outdir, exist_ok=True)
         outfilename = os.path.join(outdir, filename)
         if os.path.isfile(outfilename):
             if clobber:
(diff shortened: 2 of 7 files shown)


bdsf/cleanup.py
```diff
@@ -34,7 +34,7 @@ def __call__(self, img):
 
             from math import log10
             bdir = img.basedir + '/misc/'
-            if not os.path.isdir(bdir): os.makedirs(bdir)
+            if not os.path.isdir(bdir): os.makedirs(bdir, exist_ok=True)
```
gmloose:
```suggestion
            os.makedirs(bdir, exist_ok=True)
```
I don't think the `if` statement makes any sense. It is similar to `os.path.exists`, but would evaluate to `True` if `bdir` were, e.g., a file. Suggest to remove it.

bdsf/functions.py
```diff
@@ -1393,8 +1392,7 @@ def write_image_to_file(use, filename, image, img, outdir=None,
         # Write image to FITS file
         if outdir is None:
             outdir = img.indir
-        if not os.path.exists(outdir) and outdir != '':
-            os.makedirs(outdir)
+        os.makedirs(outdir, exist_ok=True)
```
gmloose, line 1395: In theory `outdir` could be an empty string. I don't like the original code. I think it would be better to change line 1393 to
```
        if not outdir:
```
Then your line 1395 is OK.

bdsf/output.py
```diff
@@ -32,8 +31,7 @@ def __call__(self, img):
                 self.write_ds9(img, dir, objtype='srl')
                 self.write_gaul_FITS(img, dir)
                 self.write_srl_FITS(img, dir)
-            if not os.path.exists(img.basedir + '/misc/'):
-                os.makedirs(img.basedir + '/misc/')
+            os.makedirs(img.basedir + '/misc/', exist_ok=True)
```
gmloose, line 34: Preferably, paths should be constructed using `os.path.join`, but I'm OK to keep the original code.

gmloose (approved):
Looks good to me. Just a few nit-picks.

bdsf/cleanup.py
Siphlygon: I agree and was also confused about its role, so I decided to err on the cautious side and keep the original behaviour. Happy to remove too.

bdsf/functions.py
Siphlygon: Noted and agreed

bdsf/output.py
Siphlygon: Appreciate the note, and you're right about `os.path.join` being more robust. Just did a quick check -- there's many other places, including in `output.py` itself, where concatenation is used for paths, so I'll leave it out of scope for this fix. Happy to create another issue + PR if you think it's worth standardising project-wise

gmloose: Fine with me. 😄
But I can imagine your time is precious enough to not spend it on trivialities like these.

Siphlygon: I was actually quite happy to, but I've discovered far more that needs to be done in my own repos. If I ever get bored again and have little better to do, maybe I will open that PR! 😆

PR #360, opened 2026-07-18
Add mask support for `tresholdy.py`

AlexKurek:
The statistics now should now be more reliable if the image contains NaNs.


gmloose (approved):
LGTM.

PR #362, opened 2026-07-19
Fix `nmax` calculation in shapelets by removing incorrect upper limit

AlexKurek:
`min(max(X, A), B)` works as follows:
First, `max(X, 10)` ensures that the result is at least 10.
Then, `min(..., 10)` ensures that the result is at most 10.
As a result, `nmax` is always =10, and the previous calculations based on the island size and the beam width are discarded.

The following:
`(if nmax*nmax >= n*m: ... and if mode == 'fit': ...)`
already ensure that `nmax` does not become too large relative to the amount of data. Therefore, it is sufficient to keep only the lower bound.


gmloose (approved):
Makes sense to me.

PR #363, opened 2026-07-19
Fix SED errorbars

AlexKurek:
Presently uncertainty of the natural algorithm is calculated.
According to the rules of error propagation, the uncertainty of the common logarithm is given by:
$\Delta(\log_{10} y) = \frac{\Delta y}{y \ln 10}$.


gmloose (approved):
Yep, that's correct, AFAIK.

PR #365, opened 2026-07-19
Fix: mask pixels outside FOV before calculating statistics

AlexKurek:
The pixels that are after FOV should be masked (replaced with NaN-s) **before** calculating mean, RMS etc. Presently the masking is performed after.

### DIFF
--- a/bdsf/preprocess.py
+++ b/bdsf/preprocess.py
@@ -36,36 +36,42 @@ class Op_preprocess(Op):
             pols = ['I'] # assume I is always present
             ch0images = [img.ch0_arr]
 
-        if hasattr(img, 'rms_mask'):
-            mask = img.rms_mask
-        else:
-            mask = img.mask_arr
         opts = img.opts
 
-        # Run this before statistics to ensure pixels 
-        # outside the universe are masked before statistics are calculated
+        # Identify outside universe pixels FIRST
         if opts.check_outsideuniv:
             mylogger.userinfo(mylog, "Checking for pixels outside the universe")
             noutside_univ = self.outside_univ(img)
             img.noutside_univ = noutside_univ
-            frac_blank = round(float(noutside_univ)/float(image.shape[0]*image.shape[1]),3)
+            # Use img.ch0_arr.shape instead of undefined 'image.shape'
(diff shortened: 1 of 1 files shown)


bdsf/preprocess.py
```diff
@@ -42,6 +42,31 @@ def __call__(self, img):
...
+            mask = N.isnan(img.ch0_arr)
+            masked = mask.any()
+            img.masked = masked
+            if masked:
+                img.mask_arr = mask
+            img.blankpix = N.sum(mask)
+
+        # Define the mask that include the newly blanked pixels
+        if hasattr(img, 'rms_mask'):
+            mask = img.rms_mask
```
gmloose: I'm confused. You don't update `img.rms_mask` above, yet you re-assign `img_mask` to `mask`, thus clobbering earlier calculation results in `mask`.

AlexKurek: The existing masks were incorrectly overwritten. Thank you for spotting this.
To preserve the existing masks, I have added OR to safely merge the newly blanked NaN pixels with the existing img.mask_arr and img.rms_mask. This ensures earlier masking is not clobbered.

bdsf/preprocess.py
```diff
@@ -36,36 +36,42 @@ def __call__(self, img):
...
-            img.blankpix = N.sum(mask)
-
-        # Define the mask that include the newly blanked pixels
-        if hasattr(img, 'rms_mask'):
+            nan_mask = N.isnan(img.ch0_arr)
+            img.masked = True
+            
+            # Combine newly blanked pixels with the main mask
+            if hasattr(img, 'mask_arr') and img.mask_arr is not None:
+                img.mask_arr = img.mask_arr | nan_mask
```
gmloose: Just a minor nit-pick. I presume `img.mask_arr` is a Numpy array (or `None`). So, this could be written more concisely as:
```
                img.mask_arr |= nan_mask
```
Same comment applies to line 66.

gmloose: For the rest, I think this is correct now.

gmloose (approved):
LGTM.

Issue #366, opened 2026-07-21
IndexError in gausfit.py


gmloose:
Maik Nijhuis reported an issue with the latest PyBDSF, while running a test for SKA.
```
File "/shared/fsx1/spack/install/linux-rhel9-x86_64_v3/none-none/py-bdsf-1.14.1.20260710-6uh44pdjupsjgwe5m77bskks3mntprqe/lib/python3.11/site-packages/bdsf/multi_proc.py", line 61, in worker
    result = f(val)
             ^^^^^^
File "/shared/fsx1/spack/install/linux-rhel9-x86_64_v3/none-none/py-bdsf-1.14.1.20260710-6uh44pdjupsjgwe5m77bskks3mntprqe/lib/python3.11/site-packages/bdsf/functions.py", line 1967, in eval_func_tuple
    return f_args[0](*f_args[1:])
           ^^^^^^^^^^^^^^^^^^^^^^
File "/shared/fsx1/spack/install/linux-rhel9-x86_64_v3/none-none/py-bdsf-1.14.1.20260710-6uh44pdjupsjgwe5m77bskks3mntprqe/lib/python3.11/site-packages/bdsf/gausfit.py", line 293, in process_island
    gaul, fgaul = self.fit_island(isl, opts, img)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/shared/fsx1/spack/install/linux-rhel9-x86_64_v3/none-none/py-bdsf-1.14.1.20260710-6uh44pdjupsjgwe5m77bskks3mntprqe/lib/python3.11/site-packages/bdsf/gausfit.py", line 457, in fit_island
    t * u * fit_image[x1+1, y1+1] + (1.0-t) * u * fit_image[x1, y1+1])
            ~~~~~~~~~^^^^^^^^^^^^
[...] 
        IndexError: index 2 is out of bounds for axis 1 with size 2
```
After some investigation, we concluded that the error is triggered by #309, which changed a bare except clause into a narrower catch of `ValueError` only. The issue appears to be rare, in that it only occurred in the fifth iteration of a Rapthor run. So, it is probably an edge case, where a Gaussian is found on the edge of the image, causing the interpolation to fail (because we then go outside the image boundary. Note that this is an assumption, and has not yet been proven to be case.

Investigation of the problem was somewhat hampered by the fact that the exception and stack trace are printed on `stderr`, whereas the surrounding placeholders that help locate the message are printed to `stdout`.

[...]

PR #367, opened 2026-07-22
GEC-571 Handle IndexError in gaussfit.py

mnijhuis-tos:
Solves #366

### DIFF
--- a/bdsf/gausfit.py
+++ b/bdsf/gausfit.py
@@ -469,6 +469,13 @@ class Op_gausfit(Op):
                 tbk = sys.exc_info()[2]
                 print('Warning: Ignoring IndexError in moment analysis', e, file=sys.stderr)
                 print('Image shape: ', fit_image.shape, file=sys.stderr)
+                # Print some local variables, which may not exist, depending
+                # on when the exception was raised.
+                try:
+                    print('x1: ', x1, ' y1: ', y1, file=sys.stderr)
+                    print('t: ', t, ' u: ', u, file=sys.stderr)
+                except NameError:
+                    pass
                 print('Traceback of thread is:', file=sys.stderr)
                 print('-------------------------', file=sys.stderr)
                 traceback.print_tb(tbk, file=sys.stderr)
--- a/bdsf/gausfit.py
+++ b/bdsf/gausfit.py
@@ -468,6 +468,7 @@ class Op_gausfit(Op):
             except IndexError as e:
                 tbk = sys.exc_info()[2]
                 print('Warning: Ignoring IndexError in moment analysis', e, file=sys.stderr)
+                print('Image shape: ', fit_image.shape, file=sys.stderr)
                 print('Traceback of thread is:', file=sys.stderr)
                 print('-------------------------', file=sys.stderr)
                 traceback.print_tb(tbk, file=sys.stderr)
--- a/bdsf/gausfit.py
+++ b/bdsf/gausfit.py
@@ -13,6 +13,8 @@ provides all functionality required by the specific fitters.
 """
 from __future__ import print_function
 from __future__ import absolute_import
+import sys
+import traceback
 
 from .image import *
 from . import mylogger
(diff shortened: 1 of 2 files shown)


bdsf/gausfit.py
```diff
@@ -463,6 +465,14 @@ def fit_island(self, isl, opts, img, ngmax=None, ffimg=None, ini_gausfit=None):
                                                       isl.image, size)
             except ValueError:
                 pass
+            except IndexError as e:
+                tbk = sys.exc_info()[2]
+                print('Warning: Ignoring IndexError in moment analysis', e, file=sys.stderr)
+                print('Image shape: ', fit_image.shape, file=sys.stderr)
```
scpmw, line 471: Maybe also log `t` and `u`? I suspect they are going to be close to zero here, which means that something like
```python
                    s_peak = 0
                    eta = 1e-7
                    if 1.0 - u > eta:
                        if 1.0 - t > eta:
                            s_peak += (1.0-t) * (1.0-u) * fit_image[x1, y1]
                        if t > eta:
                            s_peak += t * (1.0-u) * fit_image[x1+1, y1]
                    if u > eta:
                        if 1.0 - t > eta:
                            s_peak += (1.0-t) * u * fit_image[x1, y1+1]
                        if t > eta:
                            s_peak += t * u * fit_image[x1+1, y1+1]
```

might be the long-term "fix"? However, until that point it's likely a good idea to at least make sure this doesn't crash...

mnijhuis-tos: Since the variables may not exist if the exception occurs before creating them, I initially didn't print them. I found a way to print x1, y1, t, and u safely now.

bdsf/gausfit.py
```diff
@@ -463,6 +465,21 @@ def fit_island(self, isl, opts, img, ngmax=None, ffimg=None, ini_gausfit=None):
...
+                # on when the exception was raised.
+                try:
+                    print('x1: ', x1, ' y1: ', y1, file=sys.stderr)
+                    print('t: ', t, ' u: ', u, file=sys.stderr)
+                except NameError:
+                    pass
+                print('Traceback of thread is:', file=sys.stderr)
+                print('-------------------------', file=sys.stderr)
+                traceback.print_tb(tbk, file=sys.stderr)
+                print('-------------------------', file=sys.stderr)
```
mmancini-skao, line 482: Please use the standard logging library here. You can also use logging.exception and it will print the stack trace.

bdsf/multi_proc.py
```diff
@@ -68,12 +68,12 @@ def worker(f, ii, chunk, out_q, err_q, lock, bar, bar_state, preserve_order=Fals
...
-            print('Traceback of thread is:')
-            print('-------------------------')
-            traceback.print_tb(tbk)
-            print('-------------------------')
+            tbk = sys.exc_info()[2]
+            print('Thread raised exception', e, file=sys.stderr)
+            print('Traceback of thread is:', file=sys.stderr)
+            print('-------------------------', file=sys.stderr)
+            traceback.print_tb(tbk, file=sys.stderr)
+            print('-------------------------', file=sys.stderr)
```
mmancini-skao, line 76: same as above use logging library

bdsf/gausfit.py
```diff
@@ -463,6 +465,21 @@ def fit_island(self, isl, opts, img, ngmax=None, ffimg=None, ini_gausfit=None):
...
             except ValueError:
                 pass
+            except IndexError as e:
+                tbk = sys.exc_info()[2]
+                print('Warning: Ignoring IndexError in moment analysis', e, file=sys.stderr)
+                print('Image shape: ', fit_image.shape, file=sys.stderr)
+                # Print some local variables, which may not exist, depending
+                # on when the exception was raised.
+                try:
+                    print('x1: ', x1, ' y1: ', y1, file=sys.stderr)
```
mmancini-skao, line 475: locals().get("x1", "undefined") works as well and avoids you a nested try catch exception

mmancini-skao (commented):
Just minor comments

bdsf/gausfit.py
gmloose: PyBDSF has its own logging system. Even though it's based on the python `logging` library, its doesn't always play nice with it. I still need to find the time to figure out how this can be improved (but it has low priority). So, I'd prefer to use the sub-optimal `print()` solution here.

gmloose:
I created a PR too (just hadn't uploaded it yesterday, because Peter suggested to wait for Maik's analysis). Instead, Maik created a PR :)
Is it OK if I close this PR and merge #368?

mnijhuis-tos:
> I created a PR too (just hadn't uploaded it yesterday, because Peter suggested to wait for Maik's analysis). Instead, Maik created a PR :) Is it OK if I close this PR and merge #368?

Yes, please go ahead. #368 is very similar to this PR and provides the same solution.

PR #368, opened 2026-07-23
Issue 366

gmloose:
This should fix the `IndexError` in the `fit_island()` method that is triggered in rare occasions, probably when a Gaussian source is found very close to the edge of the image, thus causing the interpolation to index a pixel that is just outside the image boundaries.

At the same time, this fixes a bug where part of the exception message that is printed in the `worker()` method of `multi_proc.py` was sent to `stdout` and part to `stderr`. Now all lines are sent to `stderr`.

### DIFF
--- a/bdsf/gausfit.py
+++ b/bdsf/gausfit.py
@@ -23,6 +23,7 @@ if has_pl:
 import scipy.ndimage as nd
 from . import multi_proc as mp
 import itertools
+import sys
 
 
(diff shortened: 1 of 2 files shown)


darafferty (approved):
Looks good to me — I'll let you decide which PR (this one or #367) should be merged, as I think they are mostly equivalent (though currently this one logs a bit more information).

PR #370, opened 2026-07-23
Fix pixel count mismatch between total flux and its error

AlexKurek:
`total_flux` relied on `valid_pixels`, but the flux uncertainty (`total_fluxE`) calculated its pixel count (`pixels_in_isl`) by checking for NaNs in the image only. If the background map contained NaNs, `pixels_in_isl` overestimated the actual number of pixels used for the flux.


gmloose (approved):
LGTM. Apparently, we both overlooked this when you made one of your earlier PR's in the last week 😄.

PR #376, opened 2026-07-24
Fix handling invalid slice in RMS MAD calculation

AlexKurek:
If all the pixel values are equal, this still will be 0. So add another fallback. This Inf will be processed further in `rms_mean_map`.

### DIFF
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -979,7 +979,9 @@ class Op_rmsimage(Op):
                     
                     # Protection against zero noise (e.g. all pixels have the same value)
                     if cr == 0.0:
-                        cr = np.std(valid_pixels) # final fallback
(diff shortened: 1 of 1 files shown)


gmloose (commented):
That looks sane to me, but what happens if we divide by `np.inf`?

AlexKurek:
> That looks sane to me, but what happens if we divide by `np.inf`?

np.inf will never be used in a division. When rms_mean_map detects np.inf in the map, it passes it to fill_masked_regions(), which interpolates over these. Before any division happens, 0 are converted to NaN (rms_nonzero[rms <= 0.0] = np.nan). So returning np.inf here triggers the existing fallback.

gmloose (approved):

gmloose (approved):

PR #377, opened 2026-07-24
Fix brightest pixel search for adaptive RMS map

AlexKurek:
If the image contains a NaN value, this method will return the position of the NaN pixel instead of the true location of the brightness peak.


gmloose (approved):
LGTM.

PR #378, opened 2026-07-24
Make `calculate_maps` faster

AlexKurek:
I have modified `multi_proc.py` and benchmarked `calculate_maps`.

**Replaced Manager.Queue with context.Queue**
Eliminate the overhead of routing data through the process manager.

**Adjusted output collection order**
Updated `run_tasks()` to read worker outputs from the queue before calling proc.join() to prevent potential deadlocks caused by filled OS pipe buffers.

1.5 GB FITS, 20k x 20k pixels:
before:
```
        5    1.686    0.337  257.014   51.403 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/rmsimage.py:535(calculate_maps)
```
after:
```
        5    1.534    0.307  200.967   40.193 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/rmsimage.py:535(calculate_maps)
```

The entire run is 6% faster.
The .gaul files are identical.

### DIFF
--- a/bdsf/multi_proc.py
+++ b/bdsf/multi_proc.py
@@ -236,6 +236,9 @@ def parallel_map(function, sequence, numcores=None, bar=None, weights=None):
 
     preserve_order = False
     if weights is None or numcores == size:
+        # Manually split the list to avoid O(N) memory overhead 
+        # caused by numpy.array_split(), which implicitly calls np.asarray() 
+        # on standard Python lists.
         n_each_section, extras = divmod(size, numcores)
         section_sizes = ([n_each_section + 1] * extras) + ([n_each_section] * (numcores - extras))
         
--- a/bdsf/multi_proc.py
+++ b/bdsf/multi_proc.py
@@ -50,12 +50,11 @@ def worker(f, ii, chunk, out_q, err_q, lock, bar, bar_state, preserve_order=Fals
     :param f  : callable function that accepts argument from iterable
     :param ii  : process ID
     :param chunk: slice of input iterable
-    :param out_q: thread-safe output queue
-    :param err_q: thread-safe queue to populate on exception
-    :param lock : thread-safe lock to protect a resource
-           ( useful in extending parallel_map() )
+    :param out_q: process-safe output queue
+    :param err_q: process-safe queue to populate on exception
+    :param lock : lock object (kept for signature compatibility)
     :param bar: statusbar to update during fit
-    :param bar_state: statusbar state dictionary
+    :param bar_state: dictionary holding shared memory Values for statusbar state
     :param preserve_order: whether chunk entries carry their original index
     """
     vals = []
@@ -82,16 +81,17 @@ def worker(f, ii, chunk, out_q, err_q, lock, bar, bar_state, preserve_order=Fals
             vals.append(result)
 
         if bar is not None:
-            if bar_state['started']:
-                bar.pos = bar_state['pos']
-                bar.spin_pos = bar_state['spin_pos']
-                bar.started = bar_state['started']
+            # Access shared memory values directly without manager overhead
+            if bool(bar_state['started'].value):
+                bar.pos = bar_state['pos'].value
+                bar.spin_pos = bar_state['spin_pos'].value
+                bar.started = bool(bar_state['started'].value)
                 increment = bar.increment()
-                bar_state['started'] = bar.started
-                bar_state['pos'] += increment
-                bar_state['spin_pos'] += increment
-                if bar_state['spin_pos'] >= 4:
-                    bar_state['spin_pos'] = 0
+                bar_state['started'].value = int(bar.started)
+                bar_state['pos'].value += increment
+                bar_state['spin_pos'].value += increment
+                if bar_state['spin_pos'].value >= 4:
+                    bar_state['spin_pos'].value = 0
 
     out_q.put((ii, vals))
 
@@ -102,8 +102,8 @@ def run_tasks(procs, err_q, out_q, num, preserve_order=False, total_items=None):
     the resultant array. Checks error queue for any exceptions.
 
     :param procs: list of Process objects
-    :param out_q: thread-safe output queue
     :param err_q: thread-safe queue to populate on exception
+    :param out_q: thread-safe output queue
     :param num : length of resultant array
     :param preserve_order: whether worker outputs carry original item indices
     :param total_items: total number of items to reconstruct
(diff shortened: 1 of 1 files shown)


AlexKurek:
Another 13 sec faster:
```
5    1.477    0.295  187.306   37.461 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/[rmsimage.py:535](http://rmsimage.py:535/)(calculate_maps)
```

AlexKurek:
Another 7 sec.
```
        5    1.481    0.296  180.523   36.105 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/rmsimage.py:535(calculate_maps)
```

AlexKurek:
Another 2 sec.
```
        5    1.492    0.298  178.721   35.744 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/rmsimage.py:535(calculate_maps)
```

bdsf/multi_proc.py
```diff
@@ -50,48 +51,77 @@ def worker(f, ii, chunk, out_q, err_q, lock, bar, bar_state, preserve_order=Fals
...
     :param bar: statusbar to update during fit
-    :param bar_state: statusbar state dictionary
+    :param bar_state: dictionary holding shared memory Values for statusbar state
     :param preserve_order: whether chunk entries carry their original index
     """
-    vals = []
+    # Disable the garbage collector in the worker process. 
+    # This prevents the GC from touching object headers of large arrays 
+    # inherited via fork, thus avoiding massive Copy-on-Write memory overhead.
+    gc.disable()
```
gmloose, line 64: Is this still necessary, after you've called `gc.freeze()` in the main thread? It probably won't harm, but I wonder if it gains you much.

bdsf/multi_proc.py
```diff
@@ -50,48 +51,77 @@ def worker(f, ii, chunk, out_q, err_q, lock, bar, bar_state, preserve_order=Fals
...
+            try:
+                result = f(val)
+            except Exception as e:
+                print("Thread raised exception", e, file=sys.stderr)
+                print("Traceback of thread is:", file=sys.stderr)
+                print("-------------------------", file=sys.stderr)
+                traceback.print_exception(e, file=sys.stderr)
+                print("-------------------------", file=sys.stderr)
+                err_q.put(e)
+                return
```
gmloose, line 110: The loop unswitching resulted in code duplication. Maybe add a little nested helper function, like `handle_exception()`?

bdsf/multi_proc.py
```diff
@@ -50,48 +51,77 @@ def worker(f, ii, chunk, out_q, err_q, lock, bar, bar_state, preserve_order=Fals
...
+                if bool(bar_state['started'].value):
+                    bar.pos = bar_state['pos'].value
+                    bar.spin_pos = bar_state['spin_pos'].value
+                    bar.started = bool(bar_state['started'].value)
+                    increment = bar.increment()
+                    bar_state['started'].value = int(bar.started)
+                    bar_state['pos'].value += increment
+                    bar_state['spin_pos'].value += increment
+                    if bar_state['spin_pos'].value >= 4:
+                        bar_state['spin_pos'].value = 0
```
gmloose, line 124: The loop unswitching resulted in code duplication. Maybe add a little nested helper function, like `update_bar_state()`?

bdsf/multi_proc.py
```diff
@@ -50,48 +51,77 @@ def worker(f, ii, chunk, out_q, err_q, lock, bar, bar_state, preserve_order=Fals
...
-            return
-
-        if preserve_order:
-            vals.append((val_idx, result))
-        else:
-            vals.append(result)
+    vals = []
+    
+    # Create local reference to append method for faster lookups in tight loop
+    append_val = vals.append
```
gmloose, line 69: Is this really a game-changer? I had expected that the Python interpreter would be better at this.

bdsf/multi_proc.py
gmloose: Or maybe we could lift the `try`/`except` outside the `for`-loop. But that has the disadvantage that we may also catch exceptions that are _not_ caused by calling `f(val)`.

bdsf/multi_proc.py
```diff
@@ -116,6 +146,12 @@ def run_tasks(procs, err_q, out_q, num, preserve_order=False, total_items=None):
         for proc in procs:
             proc.start()
 
+        # Retrieve results from queue BEFORE joining processes to avoid
+        # IPC pipe buffer deadlocks
+        raw_results = [None] * num
+        for i in range(num):
+            raw_results[i] = out_q.get()
```
gmloose, line 153: Doing a blocking read of `num` results from the queue could result in a dead-lock if one or more of the workers died, e.g., due to a raised exception. That's why the processes are joined first, and after that the queue is read. See commit 6fc72bc2ecfb7e5557761e59e97b3dfab47dca33. Are the results you're reading back that large that there's an actual risk of creating a dead-lock due to a full pipe buffer?

bdsf/multi_proc.py
```diff
@@ -191,23 +224,29 @@ def parallel_map(function, sequence, numcores=None, bar=None, weights=None):
...
     if weights is None or numcores == size:
-        sequence = numpy.array_split(sequence, numcores)
+        n_each_section, extras = divmod(size, numcores)
+        section_sizes = ([n_each_section + 1] * extras) + ([n_each_section] * (numcores - extras))
+        
+        sequence = []
+        start = 0
+        for sec_size in section_sizes:
+            sequence.append(sequence_list[start:start + sec_size])
+            start += sec_size
```
gmloose, line 249: It's not completely clear to me why you chose to remove the `numpy.array_split()` call and do all the heavy lifting of splitting a normal python list exactly the same way that `numpy.array_split()` does. Is that only to avoid the overhead of creating the Numpy array?

gmloose (commented):
See my comments.

bdsf/multi_proc.py
AlexKurek: I think its better to leave it there. i have modified the comment.

bdsf/multi_proc.py
AlexKurek: I agree. reversed that.

bdsf/multi_proc.py
AlexKurek: I have added a comment here: https://github.com/lofar-astron/PyBDSF/pull/378/commits/6f95f47bb7e4e647b5192c70e4d6bc14f242b193.

PR #379, opened 2026-07-28
Fix flux and RMS calculation in plotting

AlexKurek:
**Bug**
The code fetched boundaries via `ax.get_xlim()` and `ax.get_ylim()`. In the case of images with reversed axes (such as RA orientation in astronomy, which increases to the left), zooming in caused these functions to return the boundaries in reverse order (xmin > xmax). Slicing a NumPy array this way (e.g., img_ch0[100:20]) doesn't throw an error, but it returns an empty array. In result, the flux and RMS for the zoomed region were 0.0.

**Fix**
The limits are now sorted first.

### DIFF
--- a/bdsf/plotresults.py
+++ b/bdsf/plotresults.py
@@ -556,21 +556,18 @@ def on_press(event):
                 break
         if not has_image:
             return
-        if xmin < 0:
-            xmin = 0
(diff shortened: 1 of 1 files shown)


gmloose (approved):
LGTM.

PR #380, opened 2026-07-28
Fix shapelets statistics

AlexKurek:
**Problem**
Shapelet residual statistics returned NaN because valid pixel indices were selected using the original image (img.ch0_arr) rather than resid_shap. This allowed newly masked NaN values in resid_shap to pass into the statistical numpy functions.

**Fix**
Replaced img.ch0_arr indexing with direct NaN filtering on the shapelet residual array itself using `resid_shap[~N.isnan(resid_shap)]`.


gmloose (approved):
LGTM.

PR #381, opened 2026-08-01
Fix `jmax` calcultion in wavelets

AlexKurek:
The behavior of `/` changed in Python 3, now `//` should be used there. Also casting to int was not necessary.

### DIFF
--- a/bdsf/wavelet_atrous.py
+++ b/bdsf/wavelet_atrous.py
@@ -85,7 +85,7 @@ class Op_wavelet_atrous(Op):
                     min_size = min(max_isl_shape) * 4.0
                 else:
                     min_size = min(resid.shape)
-                jmax = int(floor(log((min_size / 3.0 * 3.0 - l) / (l - 1) + 1) / log(2.0) + 1.0)) + 1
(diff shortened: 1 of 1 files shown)


bdsf/wavelet_atrous.py
```diff
@@ -85,7 +85,7 @@ def __call__(self, img):
                     min_size = min(max_isl_shape) * 4.0
                 else:
                     min_size = min(resid.shape)
-                jmax = int(floor(log((min_size / 3.0 * 3.0 - l) / (l - 1) + 1) / log(2.0) + 1.0)) + 1
+                jmax = floor(log(((min_size // 3) * 3.0 - l) / (l - 1) + 1) / log(2.0) + 1.0) + 1
```
gmloose, line 88: I guess the intention of the original code was to do an integer division by `3`, and not a (floating point) division by `3.0`, because `min_size / 3.0 * 3.0` is equivalent to `float(min_size)`. So, I guess your fix not only makes this Python3 compatible, but also fixes the original bug.

gmloose (approved):

PR #382, opened 2026-08-02
Fix `ptbin` division in `psfvary.py`

AlexKurek:
In Python 3 the operator here must be `//`.
The test failed, but the results should be better now.

### DIFF
--- a/bdsf/psf_vary.py
+++ b/bdsf/psf_vary.py
@@ -377,12 +377,12 @@ class Op_psf_vary(Op):
         if num <= 100: ptpbin = num // 5
         if num > 100: ptpbin = num // 10
         if num > 1000: ptpbin = num // 20
-        if ptpbin % 2 == 1: ptpbin = ptpbin+1
-        if num < 10: ptpbin=num
+        if ptpbin % 2 == 1: ptpbin = ptpbin + 1
+        if num < 10: ptpbin = num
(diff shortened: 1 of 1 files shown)


bdsf/psf_vary.py
```diff
@@ -372,17 +372,17 @@ def trans_gaul(self, q):
...
+        if num > 1000: ptpbin = num // 20
+        if ptpbin % 2 == 1: ptpbin = ptpbin+1
         if num < 10: ptpbin=num
         ptpbin = float(ptpbin) # cast to float to avoid integer division errors
-        nbin=int((num-ptpbin)/(ptpbin/over)+1)
-        ptplastbin=int((num-1)-(nbin-1)*ptpbin/over)
-        nbin=nbin+1
+        nbin = int((num-ptpbin)/(ptpbin/over)+1)
+        ptplastbin = int((num-1)-(nbin-1)*ptpbin/over)
+        nbin = nbin+1
```
gmloose: Since you're adding white space for readability (hooray!), please also add some white space to the RHS expression.

gmloose:
@darafferty, could you double check, and if you agree update the reference outputs?

darafferty:
Yep, I agree with the fix. I'll make an PR to update the reference files once this one is merged (or is there some way to do that in this PR?).

darafferty (approved):

PR #386, opened 2026-08-02
Fix unintended mutation of `img.sources`

AlexKurek:
**Bug**
This was only creating a reference so the original list was modified.

**Fix**
Changed `+=` to standard list concatenation (+) to ensure a new list is created. This preserves the original state of the `img.sources`.


gmloose (approved):
Good catch! This one is not obvious to spot.

PR #387, opened 2026-08-04
Replace `IOError` with `OSError` in `read_image_from_file`

AlexKurek:
This is recommended from Python 3.3 (PEP 3151).


gmloose (approved):
LGTM. You can see that this code is really showing its age 😄

PR #388, opened 2026-08-06
Fix PA angle normalization in `deconv2`

AlexKurek:
This appears to be a precedence bug. The intended expression is:

`(gaus_c[2] + 900.0) % 180.0`

The 900 (5*180) is a legacy from Fortran, where MOD() could return negative values. In Python, % already normalizes negative angles into [0, 180), so `% 180.0` is sufficient.


gmloose (approved):
LGTM.

PR #389, opened 2026-08-07
Fix masked pixels included in image moment calculations

AlexKurek:
Due to incorrect indentation, m2 (centroid) and m3 (variance) are accumulated outside the `if not mask[i]` block. So masked pixels are included in the numerators while being excluded from the denominator (m1). This leads to biased estimates of the centroid and size.


gmloose (approved):
Good catch.

PR #390, opened 2026-08-07
Fix: the sum of squared residuals is computed only over unmasked pixels

AlexKurek:
The sum of squared residuals `dev[ind]` is computed only over unmasked pixels (`ind`). However, the denominator uses len(fit), which is the total number of pixels. As a result, when a mask is present, χ² is artificially underestimated.


AlexKurek:
Im sorry, this 2 methods are about to be removed by https://github.com/lofar-astron/PyBDSF/pull/333.

PR #392, opened 2026-08-08
Fix axis boundary checking in `same_island_min`

AlexKurek:
`pix1[1]` (Y-axis) was incorrectly compared against `subn` (X-axis length) instead of `subm` (Y-axis length). Additionally, X-axis boundary checks were missing for both `pix1` and `pix2`.


gmloose (approved):
Looks correct. But this code makes my head spin 😵‍💫.

PR #393, opened 2026-08-09
Fix incorrect coordinate shift in `process_Multiple`

AlexKurek:
After successfully fitting a 2D Gaussian to the sub-image, the code attempts to map the local fitted coordinates (`para[1:3]`) back to the global image.

**Bug**
`s_imsize` contains the X and Y sizes of the bounding box. `N.sum(s_imsize)` sums these two dimensions. Taking half of this sum and applying it as a uniform offset to *both* X and Y incorrectly assumes the sub-image is always square.

**Fix**
The correct offset from a sub-image to the parent image is simply the bottom left corner index (`blc`).

### DIFF
--- a/bdsf/gaul2srl.py
+++ b/bdsf/gaul2srl.py
@@ -353,7 +353,7 @@ class Op_gaul2srl(Op):
                 maxpeak = para[0]
             else:
                 maxpeak = maxv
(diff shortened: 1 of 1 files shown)


gmloose (approved):
Your reasoning looks sound.

PR #394, opened 2026-08-09
Fix Monte Carlo variance bias in `process_Multiple`

AlexKurek:
If the moment analysis fails for a specific iteration, the code forces the parameter to the original deterministic value.

**Bug**
By replacing failed iterations with the exact mean value, the code compresses the distribution. When `N.std(mompara0_MC)` is calculated later, this padding biases the standard deviation toward 0.


gmloose (approved):
LGTM.

PR #395, opened 2026-08-09
Fix flux error aggregation for multi-component sources

AlexKurek:
`totE` (total flux error) in multi-Gaussian sources was underestimated. Aggregating errors in quadrature assumes independent Gaussian components, which is invalid for overlapping fits within a single island.

**Fix**
Replaced root-sum-square addition with linear summation to account for component covariance (in the absence of a full covariance matrix).

### DIFF
--- a/bdsf/gaul2srl.py
+++ b/bdsf/gaul2srl.py
@@ -444,11 +444,10 @@ class Op_gaul2srl(Op):
 
         # update all objects etc
         tot = 0.0
-        totE_sq = 0.0
+        totE = 0.0
         for g in g_sublist:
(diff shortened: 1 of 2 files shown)


gmloose (commented):
This is probably correct, but I would like @darafferty's opinion on this too.

darafferty:
Yes, I think this change makes sense, but please also update the text regarding the errors in the documentation (https://github.com/lofar-astron/PyBDSF/blob/master/doc/source/algorithms.rst?plain=1#L63).

AlexKurek:
thank you for pointing this out. Done.

PR #396, opened 2026-08-09
Fix stability of MC error estimation in `gaul2srl.py`

AlexKurek:
20 iterations in MC is a very low number. It was defined at least 15y ago according to BLAME.  It would be better to use 1k (2% variance), but for a compromise, maybe 200 is enough.

The variance of the estimator drops from ~16.2% to ~5.0%.

### DIFF
--- a/bdsf/gaul2srl.py
+++ b/bdsf/gaul2srl.py
@@ -463,7 +463,7 @@ class Op_gaul2srl(Op):
         errors = func.get_errors(img, plist, isl.rms)
 
         if img.opts.do_mc_errors:
-            nMC = 20
+            nMC = 200
             mompara0_MC = N.zeros(nMC, dtype=N.float32)
             mompara1_MC = N.zeros(nMC, dtype=N.float32)
             mompara2_MC = N.zeros(nMC, dtype=N.float32)


bdsf/gaul2srl.py
```diff
@@ -463,7 +463,7 @@ def process_Multiple(self, img, g_sublist, mask, src_index, isrc, subim, isl, de
         errors = func.get_errors(img, plist, isl.rms)
 
         if img.opts.do_mc_errors:
-            nMC = 20
+            nMC = 200
```
gmloose, line 466: Since you're very keen on measuring performance, I guess you checked what the impact is of increasing this number?

AlexKurek: I can see it in cProfile, its 0.2 sec. longer.

gmloose: In absolute terms that's almost nothing. What about relative terms? What is the total run time?

AlexKurek: All right, I should have written this. The total time is 16 minutes and it's a 20k x 20k image.

gmloose (approved):

PR #397, opened 2026-08-11
Fix underestimation of `deconv_size_skyE`

AlexKurek:
This error was simply copied from `size_skyE`, but deconvolution always introduces new uncertainty.


gmloose (commented):
This is too much outside of my domain knowledge. I'll leave @darafferty to judge.

PR #400, opened 2026-08-15
Fix incorrect amplitude initialization in `fit_mask_1d`

AlexKurek:
When setting initial parameters for the spectral index model (`sp_in`) with one positive flux data point (`len(ind1) == 1`), the amplitude parameter (`p0[0]`) was mistakenly initialized using the array index (`ind1[0]`) instead of a flux value.

### DIFF
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -424,18 +424,20 @@ def fit_mask_1d(x, y, sig, mask, funct, do_err, order=0, p0 = None):
                 p0=N.array([yfit[N.argmax(xfit)]] + [1.])
             if funct == sp_in:
                 ind1 = N.where(yfit > 0.)[0]
-                if len(ind1) >= 2:
-                    low = ind1[0]; hi = ind1[-1]
-                    sp = N.log(yfit[low]/yfit[hi])/N.log(xfit[low]/xfit[hi])
-                    p0=N.array([yfit[low]/pow(xfit[low], sp), sp] + [0.]*(order-1))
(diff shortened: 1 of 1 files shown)


bdsf/functions.py
```diff
@@ -419,7 +419,11 @@ def fit_mask_1d(x, y, sig, mask, funct, do_err, order=0, p0 = None):
                     sp = N.log(yfit[low]/yfit[hi])/N.log(xfit[low]/xfit[hi])
                     p0=N.array([yfit[low]/pow(xfit[low], sp), sp] + [0.]*(order-1))
                 elif len(ind1) == 1:
-                    p0=N.array([ind1[0], -0.8] + [0.]*(order-1))
+                    idx = ind1[0]
+                    x0 = xfit[idx]
+                    y0 = yfit[idx]
+                    alpha0 = -0.8
```
gmloose: Why not use `sp` here as well, to be consistent in naming variables.
Also, I wonder if it wouldn't make sense to only put the calculation of `sp` in the `len()` conditional, to avoid code duplication. The rest is identical; I mean `low` and `high` will be the same if `len(ind1) == 1`, but that's intended, right? And the calculation of `p0` is also the same, right?

AlexKurek: I agree, done. We just need to keep the calculation of sp inside the conditional to avoid a division by 0 (log(1)/log(1)) when low == hi in the 1-point case.

gmloose (approved):

PR #401, opened 2026-08-15
Fix crash due to typo in `fit_mulgaus2d`

AlexKurek:
Simply a typo.


gmloose (approved):
Please note that `ind` in lines 939-940 and `p_tofix` will be undefined as well if `ngaus` is `0` or less (thanks to my IDE).

PR #402, opened 2026-08-15
Fix shifting only along axis 0 in `imageshift`

AlexKurek:
The function is documented to shift a 2D image in both X and Y directions using a `shift` tuple. However, passing `axis=0` to `ndimage.fourier_shift` restricts the operation to a single axis.

### DIFF
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -508,7 +508,7 @@ def imageshift(image, shift):
     f1=scipy.fft.fft(image, shape[0], axis=0)
     f2=scipy.fft.fft(f1, shape[1], axis=1)
 
-    s=ndimage.fourier_shift(f2,shift, axis=0)
+    s=ndimage.fourier_shift(f2, shift)
 
     y1=scipy.fft.ifft(s, shape[1], axis=1)
     y2=scipy.fft.ifft(y1, shape[0], axis=0)
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -501,22 +501,35 @@ def std(y):
         return s*sqrt(float(l)/(l-1))
 
 def imageshift(image, shift):
-    """ Shifts a 2d-image by the tuple (shift). Positive shift is to the right and upwards.
-    This is done by fourier shifting. """
-    import scipy.fft
-    from scipy import ndimage
+    """
+    Shift a 2D image using Fourier phase shifts.
 
-    shape=image.shape
+    Parameters
+    ----------
+    image : 2D array_like
+        Input image array.
+    shift : tuple or list of float
+        Shift values in pixels along (y, x) axes. Positive y shifts
+        the image upwards, and positive x shifts it to the right.
 
-    f1=scipy.fft.fft(image, shape[0], axis=0)
-    f2=scipy.fft.fft(f1, shape[1], axis=1)
+    Returns
+    -------
+    shifted_image : ndarray
+        Shifted image (real values).
+    """
 
-    s=ndimage.fourier_shift(f2, shift)
+    from scipy.fft import fft2, ifft2
+    from scipy.ndimage import fourier_shift
 
-    y1=scipy.fft.ifft(s, shape[1], axis=1)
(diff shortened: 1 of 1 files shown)


bdsf/functions.py
```diff
@@ -508,7 +508,7 @@ def imageshift(image, shift):
     f1=scipy.fft.fft(image, shape[0], axis=0)
     f2=scipy.fft.fft(f1, shape[1], axis=1)
 
-    s=ndimage.fourier_shift(f2,shift, axis=0)
+    s=ndimage.fourier_shift(f2, shift)
```
gmloose: Agreed. However, there's yet another discrepancy between the doc-string and the code, because positive `y` will shift downwards, instead of upwards.

Also, the two separate 1D FFT calls could be coalesced into a single `scipy.fft.fft2(image)` call; similarly for the inverse FFTs.

AlexKurek: Good catch, thank you. I have also slightly refactored this function (docstring etc.)

gmloose: I don't know why the tests now fail. Maybe the result of `fft2` is different from the two subsequent calls of `fft`? I lack the domain knowledge to judge whether we're good to go or not. Weirdly enough, I only see a single check has been done. I would have expected at least two pipeline runs: one with your initial change - the removal of `axis=0`; and one with the use of `fft2`. Then we could have compared the results.

gmloose: Oh, found it: https://github.com/lofar-astron/PyBDSF/actions/runs/31894486693/job/95035559414. That one passed. Don't know why this page says only one check has been done.

AlexKurek: The check mail have failed due to the reversal of the y-shift sign.

gmloose: Maybe the failure is caused by the fact that the image is now flipped? That shouldn't have a major effect on the statistics that are used to compare the images. But, maybe we should not do the y-flip and change the doc string instead?

AlexKurek: Maybe. This method is only called from `stackpsf` in functions.py. I will look into this later - we have people from ASTRON at our station right now, converting it to LOFAR 2.0 :)

gmloose: Yep. I can confirm that by removing the flip, the test passes again. Also, the differences in the output _with_ the flip are quite significant:
```
ERROR [root]: Image files 'tbdsf_process_image.in_fits_pybdsf/misc/tbdsf_process_image.in_fits.pybdsf.psf_vary_min.fits' and 'reference_outputs/misc/tbdsf_process_image.in_fits.pybdsf.psf_vary_min.fits' differ (rtol=0.03)
Noise: 4328.008 mJy/b vs 3762.214 mJy/b
ERROR [root]: Image files 'tbdsf_process_image.in_fits_pybdsf/misc/tbdsf_process_image.in_fits.pybdsf.psf_vary_pa.fits' and 'reference_outputs/misc/tbdsf_process_image.in_fits.pybdsf.psf_vary_pa.fits' differ (rtol=0.03)
Noise: 55796.816 mJy/b vs 5286.047 mJy/b
```
So, I propose to keep the `fft2` change, to remove the y-flip, and to update the doc string instead.

gmloose: And maybe even the doc-string is correct, depending on how up and down are defined for the astronomical images.

AlexKurek: Done.

bdsf/functions.py
```diff
@@ -508,22 +508,31 @@ def std(y):
...
+    Parameters
+    ----------
+    image : 2D array_like
+        Input image array.
+    shift : tuple or list of float
+        Shift values in pixels along (y, x) axes.
+
+    Returns
+    -------
+    f4.real : ndarray
```
gmloose: One minor nit-pick. I wouldn't put the name of the returned variable in the doc-string. It's irrelevant. Prefer to write something like:
```
Returns
-------
ndarray
    Shifted 2D image
```

gmloose (approved):

PR #403, opened 2026-08-16
Add comment to `drawellipse`

AlexKurek:
The function used FWHM as the ellipse semi-axes.

**Fix**
Halved the FWHM values (`size_fwhm[0]/2.0`, `size_fwhm[1]/2.0`) when calculating the semi-axes (`x1`, `y1`) to correctly draw the FWHM contour.

### DIFF
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -267,8 +267,10 @@ def drawellipse(g):
     size = [param[3], param[4], param[5]]
     size_fwhm = corrected_size(size)
     th=N.arange(0, 370, 10)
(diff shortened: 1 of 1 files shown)


AlexKurek:
This one is wrong, since the intention (possibly) is to encircle the are to the threshold level, not to FWHM level. Im sorry.
**EDIT**: Actually, best seems to add a comment here, since this could be doubtful in the future.

gmloose (approved):
LGTM. It's always good to document things that might raise questions in the future.

PR #404, opened 2026-08-16
Fix position angle normalization in `corrected_size`

AlexKurek:
`if pa < 0.0` after `% 360` is a dead code and failed to wrap 180.0° to 0.0°.

**Fix**
`pa = (bpa - 90.0) % 180.0` handles negative inputs and enforces the $[0, 180)$ interval.


gmloose (approved):
LGTM.

PR #405, opened 2026-08-18
Remove unnecessary check in `isl_tosplit`

AlexKurek:
**Bug**
Wrapping `None` in `N.array(None)` creates a NumPy array containing `None`, which defeats subsequent `is not None` checks.

**Fix**
Conditionally convert to a NumPy array only if the input is not `None`.

### DIFF
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -1804,7 +1804,8 @@ def isl_tosplit(isl, opts):
     index = 0
     n_subisl3, labels3, isl_pixs3 = open_isl(isl.mask_active, 3)
     n_subisl5, labels5, isl_pixs5 = open_isl(isl.mask_active, 5)
-    isl_pixs3, isl_pixs5 = N.array(isl_pixs3), N.array(isl_pixs5)
+    isl_pixs3 = N.array(isl_pixs3) if isl_pixs3 is not None else None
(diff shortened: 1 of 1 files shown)


bdsf/functions.py
```diff
@@ -1804,7 +1804,8 @@ def isl_tosplit(isl, opts):
     index = 0
     n_subisl3, labels3, isl_pixs3 = open_isl(isl.mask_active, 3)
     n_subisl5, labels5, isl_pixs5 = open_isl(isl.mask_active, 5)
-    isl_pixs3, isl_pixs5 = N.array(isl_pixs3), N.array(isl_pixs5)
+    isl_pixs3 = N.array(isl_pixs3) if isl_pixs3 is not None else None
+    isl_pixs5 = N.array(isl_pixs5) if isl_pixs5 is not None else None
```
gmloose: I don't understand the original line. Your fix is OK, but IMHO is defensive programming. Unless I'm missing something here, `open_isl()` is already returning either a `numpy.array` or `None`. So, I think the original line (1807) should be removed and your fix is unnecessary. But please correct me if you think I'm wrong.

AlexKurek: I agree. I have removed the unnecessary check and a few unused variables.

gmloose (approved):
LGTM.

PR #406, opened 2026-08-19
Fix `IndexError` and optimize bilinear interpolation in `fit_island`

### LINKED ISSUE #366: IndexError in gausfit.py

AlexKurek:
**Bug**
The bilinear interpolation triggered an `IndexError` at image boundaries, causing valid islands to be skipped.

**Fix**
Replaced the interpolation logic with `scipy.ndimage.map_coordinates()`. This natively handles edge boundaries.
Maybe it will also fix https://github.com/lofar-astron/PyBDSF/issues/366.

### DIFF
--- a/bdsf/gausfit.py
+++ b/bdsf/gausfit.py
@@ -449,12 +449,7 @@ class Op_gausfit(Op):
                 mompara = func.momanalmask_gaus(fit_image, mask_id, isl.island_id, pixel_beamarea, True)
                 mompara[5] += 90.0
                 if not N.isnan(mompara[1]) and not N.isnan(mompara[2]):
-                    x1 = int(N.floor(mompara[1]))
-                    y1 = int(N.floor(mompara[2]))
-                    t = (mompara[1]-x1)/(x1+1-x1)
-                    u = (mompara[2]-y1)/(y1+1-y1)
-                    s_peak = ((1.0-t) * (1.0-u) * fit_image[x1, y1] + t * (1.0-u) * fit_image[x1+1, y1] +
-                              t * u * fit_image[x1+1, y1+1] + (1.0-t) * u * fit_image[x1, y1+1])
+                    s_peak = float(nd.map_coordinates(fit_image, [[mompara[1]], [mompara[2]]], order=1, mode='nearest')[0])
                     mompara[0] = s_peak
                     par = mompara.tolist()
                     par[3] /= fwsig
@@ -464,14 +459,6 @@ class Op_gausfit(Op):
                                                       isl.image, size)
             except ValueError:
                 pass
-
-            except IndexError:
-                print("IndexError in fit_island():", file=sys.stderr)
-                try:
-                    print(f"  {fit_image.shape=}", file=sys.stderr)
-                    print(f"  {mompara=}, {x1=}, {y1=}, {t=}, {u=}", file=sys.stderr)
-                except (AttributeError, NameError):
-                    pass
 
         # Return whatever we got
         if verbose:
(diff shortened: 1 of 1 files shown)


gmloose (commented):
Funny. The handling of `IndexError` was a recent addition (commit 754be462), because in rare cases, we noticed that it was raised, and we wanted to get a grasp on what triggered it.

I don't feel comfortable enough to judge the correctness of your solution. So, I'll hand it over to @darafferty.

AlexKurek:
This is similar to https://github.com/lofar-astron/PyBDSF/pull/369 but not identical.

gmloose:
> This is similar to #369 but not identical.

Agree. I think a mix of both is (probably) the best solution. I think we should keep the `mode="nearest"`, and I think the cast to `float` is not necessary. Since this line of code is likely not executed very often, I don't think we need to worry about performance, so using `map_coordinates` is probably fine, even though the original code (if fixed to properly handle edge cases) is probably faster.

AlexKurek:
Strange, both with casting and without are crashing the test now.

PR #407, opened 2026-08-21
Fix Gaussian size error scaling in `g2param_err`

AlexKurek:
**Bug**
In `g2param_err` where `g.size_pix_adj` (containing size values in pixels) was incorrectly read instead of uncertainty values when `adj=True`.

**Fix**
Scale the actual size errors (`g.size_pixE`) by the ratio `g.size_pix_adj / g.size_pix` when `adj=True`.

### DIFF
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -218,21 +218,28 @@ def g2param(g, adj=False):
 
     return params
 
+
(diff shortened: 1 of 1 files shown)


gmloose (approved):
Looks sensible to me.

PR #408, opened 2026-08-22
Fix threshold index search

AlexKurek:
Because `sc.s` is sorted in ascending order, the original `s < smin_L` condition broke immediately on the first element, locking `index` permanently to `0`.


gmloose (approved):
Good catch.

AlexKurek:
I'm still not sure if this method based on sourcecount.py is working this days. This must have been done on some 1.4 GHz catalogue where (AFAIK) galaxies looks much different than in LOFAR or SKA images. But maybe after all this fixes it is a bit less not working;)

PR #410, opened 2026-08-23
Fix corruption of an island mask in `assign_leftovers`

AlexKurek:
**Mask updating**
Replaced the incorrect reassignment (`mask = (mlabels == ii)`) with proper in-place pixel masking (`mask[tuple(cc)] = True`). (There was a commented-out attempt `# mask[cc] = True` indicating the original intent).

**Loop termination**
The return was executed too early, meaning that not all pixels could be processed.

### DIFF
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -1734,6 +1734,7 @@ def assign_leftovers(mask, open, nisl, labels):
             for cc in coords:
                 # Convert 'cc' to a tuple to properly index the 2D numpy array
                 mask[tuple(cc)] = True
+            return labels, mask
             
         if len(belongs) == 1:
             for cc in coords:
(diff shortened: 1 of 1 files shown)


bdsf/functions.py
```diff
@@ -1728,12 +1728,13 @@ def assign_leftovers(mask, open, nisl, labels):
...
         if len(belongs) == 0:
-            # No suitable islands found => mask pixels
+            # No suitable islands found, so mask individual pixels
             for cc in coords:
-                mask = (mlabels == ii)
-#             mask[cc] = True
-                return None, mask
+                # Convert 'cc' to a tuple to properly index the 2D numpy array
+                mask[tuple(cc)] = True
+
```
gmloose, line 1461: You removed the `return` statement, which changes the logic of the function. It will now continue into the `else` branch of `if len(belongs) == 1`, which is probably not what you want. OTOH, I find the original method confusing in the sense that it either returns `None` or a `list`. I'm not sure why it doesn't simply return the unaltered `labels` in this case, which I think it should. @darafferty, what do you think?

AlexKurek: I agree, I have restored `return`.

gmloose: Yeah, well the question remains if it should return `labels`, which is provided as input and will most likely _not_ be `None`, or `None` as it was. Like I said, I'm confused about the original code, so this might as well be a bug fix.

PR #411, opened 2026-08-25
Fix NaNs in island/source residual statistics in `calc_resid_mean_rms`

AlexKurek:
Global masks `img.rms_mask` or `img.mask_arr` inject NaNs into the residual images. These NaNs occasionally bypass the island-level masks (~isl.mask_active), causing N.std() and N.mean() to fail.


gmloose (approved):
LGTM. Using the NaN versions of `std` and `mean` should be harmless and yield more accurate statistics.

PR #412, opened 2026-08-25
Fix invalid IDs assigned to flagged Gaussians in `fgaul`

AlexKurek:
`gaul[-1][0]` was used to find the last Gaussian index, but it actually retrieved the peak flux. Due to Python 3 comprehension scoping, `gidx` retained this value, causing flagged Gaussians to receive invalid IDs.

**Fix**
Used `len(gaul)` as the starting index for `fgaul` items

### DIFF
--- a/bdsf/gausfit.py
+++ b/bdsf/gausfit.py
@@ -156,10 +156,6 @@ class Op_gausfit(Op):
             gaul = gaus_list[idx][0]
             fgaul = gaus_list[idx][1]
             dgaul = []
-            if len(gaul) > 0:
-                gidx = gaul[-1][0]  # save last index value for use with fgaul below
-            else:
-                gidx = 0
             gaul = [Gaussian(img, par, idx, gidx)
                     for (gidx, par) in enumerate(gaul)]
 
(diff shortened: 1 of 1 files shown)


gmloose:
I don't completely understand the issue. The reuse of `gidx` is confusing, to say the least. But I don't understand your proposed change to line 185. What would be wrong with the following?
```
last_gidx = gaul[-1][0] if gaul else 0

gaul = [
    Gaussian(img, par, idx, gidx)
    for gidx, par in enumerate(gaul)
]

...

fgaul = [
    Gaussian(img, par, idx, last_gidx + gidx + 1, flag)
    for gidx, (flag, par) in enumerate(fgaul)
]
```

AlexKurek:
> save last index value... 

was misleading about what gaul[-1][0] contains.

`gaul` is a list of parameters [amp, x, y, s1, s2, pa]. Therefore, gaul[-1][0] retrieves the peak flux of the last Gaussian, not an integer index.

In Python 2, list comprehensions leaked their loop variables, so `gidx` was accidentally overwritten by `enumerate(gaul)` to `len(gaul) - 1`. In Python 3, list comprehensions have isolated scope, so `gidx` retained the peak flux from gaul[-1][0]. For me it looks as an issue of Python 2 vs 3 and I should have mentioned it, Im sorry.

`last_gidx = gaul[-1][0]` would still store the peak flux rather than an index. Since gaul items receive indices `0` through `len(gaul) - 1`, using `len(gaul)` as the starting index for `fgaul` ensures proper IDs.

PR #413, opened 2026-08-25
Fix detection threshold offset in `gausfit.py`

AlexKurek:
Fitting operates on background-subtracted images (`isl.image - isl.islmean`), but detection thresholds (`thr`) included `isl.mean`. This inflated thresholds, causing premature fitting termination.

This error should not be large though (vide https://github.com/lofar-astron/PyBDSF/pull/350).

### DIFF
--- a/bdsf/gausfit.py
+++ b/bdsf/gausfit.py
@@ -336,7 +336,7 @@ class Op_gausfit(Op):
         if abs(beam[0]/beam[1]) < 1.1:
             beam = (1.1*beam[0], beam[1], beam[2])
 
-        thr1 = isl.mean + opts.thresh_isl*isl.rms
+        thr1 = opts.thresh_isl*isl.rms
         thr0 = thr1
         g3_only = opts.fix_to_beam
         peak = fcn.find_peak()[0]
@@ -501,7 +501,7 @@ class Op_gausfit(Op):
             opts = img.opts
         thresh_isl = opts.thresh_isl
         thresh = opts.fittedimage_clip
-        thr = isl.mean + thresh_isl * isl.rms
+        thr = thresh_isl * isl.rms
 
         if opts.verbose_fitting:
             print('Iteratively fitting island ', isl.island_id)
@@ -580,7 +580,7 @@ class Op_gausfit(Op):
         while domore:
             domore = False
             av, stdnew, maxv, maxp, minv, minp = func.arrstatmask(im1, mask)
-            if stdnew > isl.rms and maxv >= thr and maxv >= isl.mean+2.0*isl.rms:
+            if stdnew > isl.rms and maxv >= thr and maxv >= 2.0*isl.rms:
                 domore = True
                 x1, y1 = N.array(iniposn).transpose()
                 dumr = N.sqrt((maxp[0]-x1)*(maxp[0]-x1)+(maxp[1]-y1)*(maxp[1]-y1))


bdsf/gausfit.py
```diff
@@ -336,7 +336,7 @@ def fit_island(self, isl, opts, img, ngmax=None, ffimg=None, ini_gausfit=None):
         if abs(beam[0]/beam[1]) < 1.1:
             beam = (1.1*beam[0], beam[1], beam[2])
 
-        thr1 = isl.mean + opts.thresh_isl*isl.rms
+        thr1 = opts.thresh_isl*isl.rms
```
gmloose, line 339: I may be missing something, but `thr1` is only used for setting `thr0`. 🤔 
Also, I'm confused by the use of `thr0` in the call to `self.inigaus_fbdsm()`, because _inside_ that function there's this line (line 569):
```
            thr_pos = thr 
```
which treats `thr` not as a threshold, but as a position (at least, that's what the variable name suggests).

bdsf/gausfit.py
```diff
@@ -501,7 +501,7 @@ def fit_island_iteratively(self, img, isl, iter_ngmax=5, opts=None):
             opts = img.opts
         thresh_isl = opts.thresh_isl
         thresh = opts.fittedimage_clip
-        thr = isl.mean + thresh_isl * isl.rms
+        thr = thresh_isl * isl.rms
```
gmloose, line 504: Here, I think your assessment may be correct, though I'm confused about the difference between `isl.mean` and `isl.islmean` (as used in line 511 to set `peak_val`).

bdsf/gausfit.py
```diff
@@ -580,7 +580,7 @@ def inigaus_fbdsm(self, isl, thr, beam, img):
         while domore:
             domore = False
             av, stdnew, maxv, maxp, minv, minp = func.arrstatmask(im1, mask)
-            if stdnew > isl.rms and maxv >= thr and maxv >= isl.mean+2.0*isl.rms:
+            if stdnew > isl.rms and maxv >= thr and maxv >= 2.0*isl.rms:
```
gmloose, line 583: Here, I don't understand why you would have to remove `isl.mean`. The `arrstatmask` function is returning an absolute maximum, not one that has its mean subtracted.

bdsf/gausfit.py
AlexKurek: `thr_pos` stands for threshold positive (intensity threshold for positive peaks, not position).

It is passed to get_maxima() as an intensity cutoff. Since the image is background-subtracted, keeping `isl.mean` in `thr1` inflates this threshold.

bdsf/gausfit.py
AlexKurek: `isl.mean` is the background mean at the island's location, taken from the global background mean map.
`isl.islmean` is the background offset subtracted from `isl.image` for this island patch.

Line 511 uses `isl.islmean` so that `peak_val` is calculated relative to the background level that was removed from the island.

bdsf/gausfit.py
AlexKurek: `im1` is derived from the island patch (`isl.image`), which is already background-subtracted.
Therefore, `maxv` returned by `arrstatmask(im1, mask)` is already the peak height above the background (not raw flux).

AlexKurek:
Converted this to draft, since Im no logner so sure. For some time Im merging my PR-s locally like below and now I think it creates some mess. I must analyse this PR anew.

```
git clone --depth 1 https://github.com/lofar-astron/PyBDSF
cd PyBDSF

git fetch origin pull/410/head:pr410
git merge --no-edit pr410
git fetch origin pull/412/head:pr412
git merge --no-edit pr412
#git fetch origin pull/413/head:pr413
#git merge --no-edit pr413
git fetch origin pull/415/head:pr415
git merge --no-edit pr415
[...]
```

PR #415, opened 2026-08-27
Fix masked region filling bias towards 0 and loop safety in `fill_masked_regions`

AlexKurek:
- replaced zero-filling with np.nan_to_num using the global RMS median as a fallback.
- switched from np.nanmean to np.nanmedian for local region filling, making estimates robust against outliers.
- bounded the expansion search loop to image boundaries to prevent infinite execution on fully masked inputs.

### DIFF
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -934,14 +934,14 @@ class Op_rmsimage(Op):
         return cm, cr
 
 
-    def fill_masked_regions(self, themap, magic=np.inf):
+    def fill_masked_regions(self, themap):
         """Fill masked regions in themap using local median, with global median fallback."""
         # Compute global median as a fallback for finite values
-        valid_mask = np.isfinite(themap) & (themap != magic)
+        valid_mask = np.isfinite(themap)
         global_fallback = np.nanmedian(themap[valid_mask]) if np.any(valid_mask) else 1.0
 
         # Find coordinates of missing or invalid data
-        masked_boxes = np.where((themap == magic) | np.isnan(themap) | np.isinf(themap))
+        masked_boxes = np.where(~np.isfinite(themap))
         max_delx, max_dely = themap.shape[0], themap.shape[1]
 
         for i in range(np.size(masked_boxes, 1)):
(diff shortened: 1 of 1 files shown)


AlexKurek:
Generally, for filling small holes biharmonic interpolation works best (https://ui.adsabs.harvard.edu/abs/2013PASP..125.1119P/abstract) but I think it would be an overkill here.

bdsf/rmsimage.py
```diff
@@ -935,32 +935,42 @@ def process_mean_rms_maps(self, ind, mask, arr, kappa):
 
 
     def fill_masked_regions(self, themap, magic=np.inf):
```
gmloose: Having a way to override `magic=np.inf` looks a bit like over-engineering. Worse, I think it creates the possibility for subtle bugs. The only two calls I could find to this method (inside the `bdsf` package) do not pass a different value. It would simplify your code changes if we would remove the `magic` argument, and simply assume that any non-finite value should be treated as invalid.

bdsf/rmsimage.py
```diff
@@ -935,32 +935,42 @@ def process_mean_rms_maps(self, ind, mask, arr, kappa):
...
                 cutout = themap[x1:x2, y1:y2].ravel()
-                goodcutout = cutout[cutout != magic]
+                # Extract valid finite values
+                goodcutout = cutout[(cutout != magic) & np.isfinite(cutout)]
                 num_unmasked = len(goodcutout)
+
                 if num_unmasked > 0:
-                    themap[x, y] = np.nanmean(goodcutout)
+                    # Use local median from the unmasked neighborhood
+                    themap[x, y] = np.nanmedian(goodcutout)
```
gmloose, line 966: Going from `mean` to `median` changes the logic of the code. Though I agree with the argument you gave that the median is more resilient to outliers. So, I think it's fine to do here.

gmloose (commented):
LGTM, but I think we should take the opportunity so simplify this method by dropping the `magic` argument.

bdsf/rmsimage.py
AlexKurek: Good idea, done.

gmloose (approved):

PR #417, opened 2026-08-27
Fix NaN propagation for very low SNR in `for_masked_mp`

AlexKurek:
One more effort to extract a very weak signal from the noise.


gmloose (approved):
LGTM.

PR #418, opened 2026-08-27
Fix `fill_masked_regions` triggering

AlexKurek:
Ensure `fill_masked_regions` is correctly triggered when NaNs exist in the map (`NaN != np.inf` returns `True`).


bdsf/rmsimage.py
```diff
@@ -920,7 +920,7 @@ def correct_borders(map):
 
         # Step 5: fill in boxes with < 5 unmasked pixels (set to values of
         # np.inf)
-        if np.count_nonzero(mean_map != np.inf) < mapshape[0]*mapshape[1]:
+        if np.count_nonzero(np.isfinite(mean_map)) < mapshape[0]*mapshape[1]:
```
gmloose, line 922: Good, but please fix the comment above as well.

gmloose (changes requested):

PR #420, opened 2026-08-27
Fix image boundary detection in `rms_mean_map`

AlexKurek:
The subsequent `bounds` evaluated to collapse to `imgshape < imgshape`, always evaluating to `False` and silently disabling border/edge pixel processing.


gmloose (approved):
Good catch. I wonder if there are many more Python2 vs Python3 integer division issue luring around.

PR #422, opened 2026-08-30
Fix std calculation and zero division in `moment`

AlexKurek:
If variance in any dimension dropped below 0, the sqrt calculation was skipped, causing the function to return unscaled sums for all dimensions.


gmloose (approved):
LGTM.

PR #424, opened 2026-08-30
Fix NaN support in `arrstatmask`

AlexKurek:
Make it resistant for rare NaNs in arrays.

### DIFF
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -735,12 +735,12 @@ def arrstatmask(im, mask):
 
     ind = N.where(~mask)
     im1 = im[ind]
-    av = N.mean(im1)
-    std = N.std(im1)
-    maxv = N.max(im1)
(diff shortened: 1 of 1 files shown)


bdsf/functions.py
```diff
@@ -735,12 +735,12 @@ def arrstatmask(im, mask):
...
-    av = N.mean(im1)
-    std = N.std(im1)
-    maxv = N.max(im1)
+    av = N.nanmean(im1)
+    std = N.nanstd(im1)
+    maxv = N.nanmax(im1)
     x, y = N.where(im == maxv)
     xmax = x[0]; ymax = y[0]
-    minv = N.min(im1)
+    minv = N.nanmin(im1)
```
gmloose: Wouldn't it be slightly more computationally efficient if you would do the NaN filtering once, e.g., in line 736?

PR #425, opened 2026-08-30
Fix bias towards 0 for flagged Gaussians

AlexKurek:
This is simply not true, it was biasing statistics towards 0.


gmloose (approved):
LGTM.

PR #426, opened 2026-08-30
Fix flux underestimation in moment analysis

AlexKurek:
Moment analysis estimated peak brightness using bilinear interpolation on pixels around the center. Because a 2D Gaussian profile is concave near its peak, linear interpolation underestimates the peak amplitude $A$ (up to ~40%).

**Fix**
Replaced grid interpolation with an amplitude calculation: $A = F / (2\pi \sigma_x \sigma_y)$, derived from the 0th moment total flux $F$ and $\sigma$.


AlexKurek:
This one may be tricky. The idea seems obviously true and the more narrow and strong is the source, the larger the error. But looking at the tests - what I tried to do is not avalilable in the wavelets module. If we would use some other way of computing in wavelets, probably it would couse a small bias between compact and extended sources.

Also, if earlier in the code there is eg. an 'off by one' error causing that the parameters we use here for calculation do not really belong to the Gaussian they should - "fixing" this way would cause more harm than good.

PR #427, opened 2026-08-30
Add comment to `fit_island_iteratively` in `gausfit.py`

AlexKurek:
`func.corrected_size()` converts $\sigma$ to FWHM. However, `func.gaussian_fcn()` expects inputs in $\sigma$.

### DIFF
--- a/bdsf/gausfit.py
+++ b/bdsf/gausfit.py
@@ -493,7 +493,6 @@ class Op_gausfit(Op):
         and iterate. However, this does usually result in larger residuals.
         """
         from . import functions as func
-        from .const import fwsig
         sgaul = []
         sfgaul = []
         gaul = []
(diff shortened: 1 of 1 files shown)


darafferty:
Hmm, I'm confused, as it looks like the division by `fwsig` is also being done in `func.gaussian_fcn()` (in [line 706](https://github.com/lofar-astron/PyBDSF/blob/master/bdsf/functions.py#L706)). So unless I'm missing something, it seems that it does expect the FWHM?

AlexKurek:
Thats right, Im sorry, Im merging my not reviewed yet PR-s locally and probably something went wrong.

AlexKurek:
Actually, maybe better to add comment there? This is not obvious at first sight.

darafferty:
No problem — I know how confusing all these branches/PRs can be! 

Always a good idea to add comments in my opinion. PyBDSF code is poorly commented in many places unfortunately, but it's getting better thanks in no small part to your efforts!

darafferty (approved):

PR #428, opened 2026-09-05
Fix invalid `CDELT4` value in output FITS header

AlexKurek:
This is obviously not true these days. In this case maybe better to comment it out, co that full list of FITS keyword is still there.

### DIFF
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -1205,7 +1205,7 @@ def make_fits_image(imagedata, wcsobj, beam, freq, equinox, telescope, xmin=0, y
     # Add frequency info
     header['RESTFRQ'] = freq
     header['CRVAL4'] = freq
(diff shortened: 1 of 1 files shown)


gmloose (commented):
I don't understand the reasoning. AFAIK this is a standard FITS keyword. Also I don't understand your statement in the description about the full list of FITS keywords being available. Please explain.

AlexKurek:
CDELT4 stores the channel size and this line hardcodes it to 300 MHz, with is not true for LOFAR not SKA (...and looks like PyBDSF was originally written for some 1.4 GHz instrument that had 300 MHz channels).

I meant that we could remove this line simply, but maybe better to comment it out instead so that if somebody looks at the code - will see that CDELT keywords generally do not end at 3 ;) .

PR #429, opened 2026-09-06
Fix PA error calculation in `gaul2srl.py`

AlexKurek:
The std must be computed in circular space in this case.

### DIFF
--- a/bdsf/gaul2srl.py
+++ b/bdsf/gaul2srl.py
@@ -502,7 +502,26 @@ class Op_gaul2srl(Op):
             mompara4E = N.nanstd(mompara4_MC)
             if mompara4E > 2.0*mompara[4]:
                 mompara4E = 2.0*mompara[4] # Don't let errors get too large
-            mompara5E = N.nanstd(mompara5_MC)
+
+            # Use circular statistics for Position Angle error.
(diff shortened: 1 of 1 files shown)


gmloose (commented):
This is probably a good improvement, but I would like @darafferty's view on this as well.

AlexKurek:
Actually there also are:
https://docs.astropy.org/en/latest/api/astropy.stats.circstats.circstd.html#astropy.stats.circstats.circstd
and
https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.circstd.html
Probably better to use one of those.

gmloose:
> Actually there also are: https://docs.astropy.org/en/latest/api/astropy.stats.circstats.circstd.html#astropy.stats.circstats.circstd and https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.circstd.html Probably better to use one of those.

Yeah, I just stumbed over the latter one. So, maybe it's indeed better to use one of those two.

darafferty:
Ah yes, good find. Agreed it should be done with circular stats.

AlexKurek:
Done.

darafferty (approved):
Looks good to me!

PR #430, opened 2026-09-06
Make progressbar more universal

### LINKED ISSUE #385: Progress bars don't work in notebooks
The progress bar doesn't work in a Jupyter notebook (which SKA SRC users might end up needing to use as their preferred environment). Many, many errors along the lines of

stty: 'standard input': Inappropriate ioctl for device

and the bar doesn't print properly, so the output is unreadable.

`quiet=True` suppresses all the output (except some of the inappropriate ioctl errors because the bar is initialized even when it's not used).

However what would be ideal would either be to replace the progress bar with something that works in the notebook environment, as tqdm does, or else let the user disable it altogether, allowing the summary output about what's been done to remain visible.

Cosmetic but also quite easy...

AlexKurek:
This makes progressbar working when no TTY is being emulated and no longer uses ANSI character cursor positioning (e.g. `\x1b[1G` ).

There is a chance it will fix https://github.com/lofar-astron/PyBDSF/issues/385.

Tested on a GNU terminal, works fine.

### DIFF
--- a/bdsf/statusbar.py
+++ b/bdsf/statusbar.py
@@ -36,11 +36,7 @@ class StatusBar():
 
     # find number of columns in terminal
     def __getsize(self):
-        if not sys.stdout.isatty():
-            columns = 80
-        else:
-            tsize = shutil.get_terminal_size(fallback=(0, 0))
-            columns = tsize.columns
+        columns = shutil.get_terminal_size().columns
             
         if int(columns) > self.max + 2 + 44 + (len(str(self.max))*2 + 2):
             self.columns = self.max
--- a/bdsf/statusbar.py
+++ b/bdsf/statusbar.py
@@ -36,8 +36,11 @@ class StatusBar():
 
     # find number of columns in terminal
     def __getsize(self):
-        tsize = shutil.get_terminal_size(fallback=(0, 0))
-        columns = tsize.columns
+        if not sys.stdout.isatty():
+            columns = 80
+        else:
+            tsize = shutil.get_terminal_size(fallback=(0, 0))
+            columns = tsize.columns
             
         if int(columns) > self.max + 2 + 44 + (len(str(self.max))*2 + 2):
             self.columns = self.max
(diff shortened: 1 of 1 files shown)


bdsf/statusbar.py
```diff
@@ -36,8 +36,11 @@ def __init__(self, text, pos=0, max=100, color='\033[0m'):
 
     # find number of columns in terminal
     def __getsize(self):
-        tsize = shutil.get_terminal_size(fallback=(0, 0))
-        columns = tsize.columns
+        if not sys.stdout.isatty():
+            columns = 80
+        else:
+            tsize = shutil.get_terminal_size(fallback=(0, 0))
+            columns = tsize.columns
```
gmloose: Maybe we should completely rely on `shutil.get_terminal_size()` and leave out the `fallback`. By default, it will return `(80, 24)` when it cannot determine the terminal size.
```suggestion
        columns = shutil.get_terminal_size().columns
```

bdsf/statusbar.py
```diff
@@ -66,7 +69,6 @@ def __print(self):
             # so the progress bar adapts to possible window resizing.
             self.comp = int(float(self.pos) / self.max * self.columns)
             sys.stdout.write(self.color + self.text + '[' + '=' * self.comp + self.busy_char + '-'*(self.columns - self.comp - 1) + '] ' + str(self.pos) + '/' + str(self.max) + '\033[0m')
-            sys.stdout.write('\x1b[' + str(self.comp + 2 + 44) + 'G')
```
gmloose, line 69: Hm, by removing this line, you've removed a repositioning of the cursor. Not sure if this will break functionality. But you wrote that you tested this on a GNU terminal. 🤔

gmloose (commented):
I have some comments.

gmloose:
> There is a chance it will fix #385.

Not sure, there are still some places where ANSI codes are emitted. But maybe, yes.

bdsf/statusbar.py
AlexKurek: This 0, 0 looks like some kind of a fallback for headless devices. On the other hand this new condition for isatty will make it a dead code, so why not ;)

bdsf/statusbar.py
AlexKurek: I did and it looks ok. This line is strange, it seems to be moving cursor right after what was just printed, so actually not moving it.

gmloose (approved):

PR #431, opened 2026-09-06
Fix RMS explosion near strong sources

AlexKurek:
When `bstat` fails to converge within `cnt > 198`, the algorithm fell back to the unclipped mean and std (`cm = m; cr = r`). In regions containing bright sources, the unclipped std is inflated by source brightness, resulting in local RMS "explosions".

### DIFF
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -1023,14 +1023,20 @@ class Op_rmsimage(Op):
         bstat = func.bstat #_cbdsm.bstat
         a, b, c, d = ind
         if mask is None:
-            m, r, cm, cr, cnt = bstat(arr[a:b, c:d], mask, kappa)
-            if cnt > 198: cm = m; cr = r
+            _, _, cm, cr, cnt = bstat(arr[a:b, c:d], mask, kappa)
+            if cnt > 198: # use MAD
+                sub_arr = arr[a:b, c:d]
+                cm = np.nanmedian(sub_arr)
+                cr = np.nanmedian(np.abs(sub_arr - cm)) * 1.4826
         else:
             pix_unmasked = np.where(mask[a:b, c:d] == False)
             npix_unmasked = np.size(pix_unmasked,1)
             if npix_unmasked > 20: # find clipped mean/rms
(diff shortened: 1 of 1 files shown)


bdsf/rmsimage.py
```diff
@@ -1023,14 +1023,20 @@ def for_masked_mp(self, mask, arr, ind, kappa):
         bstat = func.bstat #_cbdsm.bstat
         a, b, c, d = ind
         if mask is None:
-            m, r, cm, cr, cnt = bstat(arr[a:b, c:d], mask, kappa)
-            if cnt > 198: cm = m; cr = r
+            _, _, cm, cr, cnt = bstat(arr[a:b, c:d], mask, kappa)
+            if cnt > 198: # use MAD
+                sub_arr = arr[a:b, c:d]
+                cm = np.nanmedian(sub_arr)
+                cr = np.nanmedian(np.abs(sub_arr - cm)) * 1.4826
```
gmloose: Would it be an idea to factor this out into a separate function. This code snippet now appears three times in this function (lines 1027-1030, 1035-1039, and 1044-1048); the latter with an explanation of what MAD actually is.

AlexKurek: Please note that in the first case `sub_arr = arr[a:b, c:d]`, and in two other cases MAD is computed on `arr[a:b, c:d][pix_unmasked]`, so the input is different. But I have used scipy instead on manual calculation.

gmloose (approved):
Nice!

AlexKurek:
After a fresh look: I think it is better, but still can explode. Maybe instead of median we could risk a rank operator set for some lower quartile here. Or something similar. (Median is a special case of a rank operator targetting at the middle of the set).

AlexKurek:
@gmloose Could you please revert this? I don't have the necessary permissions, but you should see a 'revert' button.

This PR may be a regression. It ignores the clipped statistics from `bstat` and uses the unclipped statistics, which should actually be worse. The efficiency of `bstat` should be optimised instead.

I'm sorry. I will review my PRs to identify any possible regressions. It may take some time...

PR #432, opened 2026-09-06
Fix false rejection of rms map

AlexKurek:
`check_rmsmap` evaluates standard deviation (`stdsub`) on the interpolated pixel array (`rms`). Interpolation acts as a low-pass filter, depressing pixel variance below the expectation (`rms_expect`). Consequently, `stdsub` frequently falls below the `1.1 * rms_expect` threshold, causing varying RMS maps to be rejected and replaced with a constant (`use_rms_map = False`).

**Fix**
Subsampled the `rms` map and `mask_arr` at step increments corresponding to the evaluation grid step size (`img.rms_box[1]`) before computing `stdsub`.

### DIFF
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -482,10 +482,15 @@ class Op_rmsimage(Op):
         cdelt = img.wcs_obj.acdelt[:2]
         bm = (img.beam[0], img.beam[1])
         fw_pix = sqrt(np.prod(bm)/abs(np.prod(cdelt)))
+        # Subsample RMS map at box step size to evaluate variance on coarse grid
+        # rather than smoothed interpolation
+        step = max(1, int(img.rms_box[1]))
         if img.masked:
-            stdsub = np.std(rms[~img.mask_arr])
+            sub_rms = rms[::step, ::step]
+            sub_mask = img.mask_arr[::step, ::step]
+            stdsub = np.std(sub_rms[~sub_mask])
         else:
-            stdsub = np.std(rms)
+            stdsub = np.std(rms[::step, ::step])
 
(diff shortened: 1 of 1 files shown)


bdsf/rmsimage.py
```diff
@@ -482,10 +482,17 @@ def check_rmsmap(self, img, rms):
         cdelt = img.wcs_obj.acdelt[:2]
         bm = (img.beam[0], img.beam[1])
         fw_pix = sqrt(np.prod(bm)/abs(np.prod(cdelt)))
+        # Subsample RMS map at box step size to evaluate variance on coarse grid
+        # rather than smoothed interpolation
+        # Determines the sampling step size in pixels (RMS box step)
+        # and enfore a minimum value to prevent a zero-step slice
```
gmloose: Typo: "enfore" -> "enforce"

gmloose (commented):
I don't understand your reasoning. Why does sub-sampling avoid interpolation? Why would calculating the standard deviation on the original data cause low-pass filtering?

AlexKurek:
`rms` array passed to `check_rmsmap` is not the original data. It has been interpolated to the full image resolution via mapcoord_threaded (3rd-order spline filter). Interpolation suppresses variance.

Because the variance is lowered, `stdsub` drops frequently below the `1.1 * rms_expect` threshold. This causes varying RMS maps to be rejected.

By sub-sampling the map at the original `rms_box[1]` step size, we are evaluating the standard deviation primarily on the nodes of the original coarse grid. This restores the true variance for the threshold check.

AlexKurek:
Im no longer so sure. I think this would require a lot of tests on a real case images. Maybe better to close this, since the users can do `rms_map = True`.

PR #359, opened 2026-07-17
Make `p2s` faster

AlexKurek:
.gaul files are identical.

before:
`   951377    3.795    0.000   17.522    0.000 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/readimage.py:176(p2s)`
after:
`   951377    1.283    0.000    8.176    0.000 /home/akurek/miniconda3/envs/pybdsf_3_14/lib/python3.14/site-packages/bdsf/readimage.py:176(p2s)`

PR #434, opened 2026-09-08
Revert "Make p2s faster"

AlexKurek:
Revert https://github.com/lofar-astron/PyBDSF/pull/359, as there is a thread-safety regression. Attaching the preallocated buffer (self._p2s_buf) to the instance creates a race condition during concurrent execution, causing threads to overwrite each other's coordinates.

Thanks to @bartoszsmierciak for noticing this.


gmloose (approved):
Oops!

PR #435, opened 2026-09-08
Partially revert "Fix negative brightsize values in rmsimage.py"

AlexKurek:
Just for clarity, the abs() here is not necessary, since the vector was already converted to abs() in readimage.py (as `acdelt`). Add comment instead.

Thanks to @BartoszSmierciak for noticing this.

### DIFF
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -187,15 +187,16 @@ class Op_rmsimage(Op):
         # Assume a 2D Gaussian beam profile: I_max * exp(-r^2 / (2*sigma^2))
         # The formula finds the diameter 2r where the brightness drops to 3x the noise level
         # (kappa1 * crms). Solving for 2r gives: 2 * sigma * sqrt(2 * ln(I_max / (3 * sigma_noise)))
+        # cdelt was already converted to abs() in readimage.py
         try:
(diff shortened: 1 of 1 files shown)


bdsf/rmsimage.py
```diff
@@ -187,15 +187,16 @@ def __call__(self, img):
         # Assume a 2D Gaussian beam profile: I_max * exp(-r^2 / (2*sigma^2))
         # The formula finds the diameter 2r where the brightness drops to 3x the noise level
         # (kappa1 * crms). Solving for 2r gives: 2 * sigma * sqrt(2 * ln(I_max / (3 * sigma_noise)))
+        # cdelt was already converted to abs() in readimage.py
         try:
-            brightsize = round(2.*img.beam[0]/abs(cdelt[0])/fwsig*
+            brightsize = round(2.*img.beam[0]/cdelt[0]/fwsig*
```
gmloose, line 192: `cdelt` is not multi-dimensional, right? In that case I wouldn't bother, and consider it defensive programming. Imagine someone removes the `abs` in `readimage.py`. Then you're doomed.

AlexKurek: Good thinking. So I will close this one.

PR #436, opened 2026-09-10
Add `adaptive_stat`

AlexKurek:
Recently, we changed many of the mean, median and std to their NaN-resistant versions. This is generally good, sice prevents rejecting a source becouse of lets say 1 NaN in the arrays, but in rare cases where the input is eg. 95% NaN, it would still be safer if the result was NaN. So I propose adding such a method.

Also, I have occasionally removed duplicate NumPy imports.

Test:

```
# 10 elements, 20% are NaNs
data = [1.0, 2.0, np.nan, 4.0, 5.0, 6.0, np.nan, 8.0, 9.0, 10.0]
print(adaptive_stat(data, statistic='mean',   threshold=30, threshold_type='percent'))
print(adaptive_stat(data, statistic='median', threshold=10, threshold_type='percent'))
print(adaptive_stat(data, statistic='std',    threshold=3,  threshold_type='count'))

# 10 elements, 60% are NaNs
data = [np.nan, 2.0, np.nan, 4.0, 5.0, 6.0, np.nan, np.nan, np.nan, np.nan]
print(adaptive_stat(data, statistic='mean',   threshold=30, threshold_type='percent'))
print(adaptive_stat(data, statistic='median', threshold=10, threshold_type='percent'))
print(adaptive_stat(data, statistic='std',    threshold=3,  threshold_type='count'))
```

> 5.625
> nan
> 3.0388114452858046
> nan
> nan
> nan

### DIFF
--- a/bdsf/functions.py
+++ b/bdsf/functions.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 from __future__ import absolute_import
 from shutil import get_terminal_size
+import numpy as N
 
 try:
     # For Python 2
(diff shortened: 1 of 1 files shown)


bdsf/functions.py
```diff
@@ -1,24 +1,74 @@
...
+    if arr.size == 0:
+        return N.nan
+
+    nan_count = N.isnan(arr).sum()
+
+    # Determine the limit depending on the threshold type
+    if threshold_type == 'percent':
+        limit = (threshold / 100.0) * arr.size
+    elif threshold_type == 'count':
+        limit = threshold
```
gmloose: You may want to cap `limit` between reasonable values, to unwanted behavior when, e.g., a value of  `-1` is passed as argument.
```suggestion
    if threshold_type == 'percent':
        limit = N.clip(threshold / 100.0, 0, 1) * arr.size
    elif threshold_type == 'count':
        limit = N.clip(int(threshold), 0, arr.size))
```

bdsf/functions.py
```diff
@@ -1,24 +1,74 @@
...
+    Parameters:
+    - data: numpy array or list (e.g., an image).
+    - statistic: string specifying the statistic to compute ('mean', 'median', or 'std').
+    - threshold: threshold value (numeric).
+    - threshold_type: 'percent' (default) or 'count' (specific number of elements).
+
+    Returns:
+    - The calculated statistic (float).
+    """
+    arr = N.asarray(data).flatten()
```
gmloose: You don't need to flatten the array. This avoid creating a copy.

bdsf/functions.py
```diff
@@ -1,24 +1,74 @@
...
 try:
     # For Python 2
     basestring = basestring
 except NameError:
     basestring = str
 
+
+def adaptive_stat(data, statistic, threshold, threshold_type='percent'):
+    """
+    Calculates a specific statistic (mean, median, or std) taking into account the number of NaNs.
```
gmloose:
```suggestion
    Calculates a specific statistic (mean, median, or std) taking into account the number of NaNs. If the number of NaNs is below the threshold, NaNs are ignored. Otherwise the result will be NaN.
```

bdsf/functions.py
```diff
@@ -1,24 +1,74 @@
...
+
+    Parameters:
+    - data: numpy array or list (e.g., an image).
+    - statistic: string specifying the statistic to compute ('mean', 'median', or 'std').
+    - threshold: threshold value (numeric).
```
gmloose:
```suggestion
    - threshold: threshold value (numeric); maximum number or percentage of NaNs that is tolerated.
```

gmloose (commented):
From a stylistic point-of-view it would be better to
```
import numpy as np
```
because this is the generally accepted way. But I would probably do that in a separate MR. Also, because this change would probably have to be made to multiple source files.

A final remark: I don't see this method being used anywhere, yet.

PR #437, opened 2026-09-10
Partially revert `Fix jmax calcultion in wavelets`

AlexKurek:
Using `(min_size // 3) * 3.0` truncates the size (e.g., 100 becomes 99). It can drop the `jmax` scale by 1, degrading the algorithm's ability to extract extended sources.

Since `... / 3.0 * 3.0` is likely an artifact with no mathematical justification, we should probably remove these redundant operations in the future.

For now I reverted this line, keeping the removal of the redundant `int()` cast.

Thanks for @Bartoszsmierciak for noticing the regression.

### DIFF
--- a/bdsf/wavelet_atrous.py
+++ b/bdsf/wavelet_atrous.py
@@ -85,7 +85,7 @@ class Op_wavelet_atrous(Op):
                     min_size = min(max_isl_shape) * 4.0
                 else:
                     min_size = min(resid.shape)
-                jmax = floor(log((min_size / 3.0 * 3.0 - l) / (l - 1) + 1) / log(2.0) + 1.0) + 1
+                jmax = floor(log((min_size - l) / (l - 1) + 1) / log(2.0) + 1.0) + 1
                 if min_size * 0.55 <= (l + (l - 1) * (2 ** (jmax) - 1)):
                     jmax = jmax - 1
             img.wavelet_lpf = lpf
--- a/bdsf/wavelet_atrous.py
+++ b/bdsf/wavelet_atrous.py
@@ -85,7 +85,7 @@ class Op_wavelet_atrous(Op):
                     min_size = min(max_isl_shape) * 4.0
                 else:
                     min_size = min(resid.shape)
-                jmax = floor(log(((min_size // 3) * 3.0 - l) / (l - 1) + 1) / log(2.0) + 1.0) + 1
+                jmax = floor(log((min_size / 3.0 * 3.0 - l) / (l - 1) + 1) / log(2.0) + 1.0) + 1
                 if min_size * 0.55 <= (l + (l - 1) * (2 ** (jmax) - 1)):
                     jmax = jmax - 1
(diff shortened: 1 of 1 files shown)


bdsf/wavelet_atrous.py
```diff
@@ -85,7 +85,7 @@ def __call__(self, img):
                     min_size = min(max_isl_shape) * 4.0
                 else:
                     min_size = min(resid.shape)
-                jmax = floor(log(((min_size // 3) * 3.0 - l) / (l - 1) + 1) / log(2.0) + 1.0) + 1
+                jmax = floor(log((min_size / 3.0 * 3.0 - l) / (l - 1) + 1) / log(2.0) + 1.0) + 1
```
gmloose, line 88: I don't fully agree. Even though the original code used `/`, that was implicitly assuming integer division, because the original code base was written for Python 2. My assumption is that the original authors intended integer division here, otherwise the expression makes no sense -- dividing by 3.0 and directly afterward multiplying by 3.0 is a no-op.

Is there a reason you want to revert this change? Do you really consider this a regression?

AlexKurek: Just not to make jmax lower, no other reason.

gmloose: I understand. But that's for the Python 3 case. But if @darafferty agrees, I'm happy to accept the revert. However, I would then remove the division and multiplication completely.
```suggestion
                jmax = floor(log((min_size - l) / (l - 1) + 1) / log(2.0) + 1.0) + 1
```

AlexKurek:
A test:

```
import math

def calculate_jmax(min_size, l, version):
    if version == 1:
        term = (min_size / 3.0 * 3.0 - l) / (l - 1) + 1
    elif version == 2:
        term = ((min_size // 3) * 3.0 - l) / (l - 1) + 1
    elif version == 3:
        term = (min_size - l) / (l - 1) + 1
        
    jmax = math.floor(math.log(term) / math.log(2.0) + 1.0) + 1
    
    # Boundary check from PyBDSF code
    if min_size * 0.55 <= (l + (l - 1) * (2 ** jmax - 1)):
        jmax -= 1
        
    return jmax

print(f"{'min_size':<10} | {'Version 1':<10} | {'Version 2':<10} | {'Version 3':<10}")
print("-" * 60)
for ms in [16, 17, 18, 64, 65, 256, 257]:
    v1, v2, v3 = [calculate_jmax(ms, 3, v) for v in (1, 2, 3)]
    print(f"{ms:<10} | {v1:<10} | {v2:<10} | {v3:<10}")
```

```
min_size   | Version 1  | Version 2  | Version 3
------------------------------------------------------------
16         | 3          | 3          | 3
17         | 4          | 3          | 4
18         | 4          | 4          | 4
64         | 5          | 5          | 5
65         | 6          | 5          | 6
256        | 7          | 7          | 7
257        | 8          | 7          | 8

```
It seems the simplified version wins.

PR #438, opened 2026-09-10
Revert `Fix RMS explosion near strong sources`

AlexKurek:
That PR was a regression. It ignored the clipped statistics from `bstat` and uses the unclipped statistics, which should actually be worse, despite MAD is used now.
The efficiency of bstat should be optimized instead. (I will extract few cases and see how well `bstat` actually works, but thats another issue).

I'm sorry. I will review my PRs to identify any possible regressions. It may take some time...

### DIFF
--- a/bdsf/rmsimage.py
+++ b/bdsf/rmsimage.py
@@ -1034,27 +1034,23 @@ class Op_rmsimage(Op):
         bstat = func.bstat #_cbdsm.bstat
         a, b, c, d = ind
         if mask is None:
-            _, _, cm, cr, cnt = bstat(arr[a:b, c:d], mask, kappa)
-            if cnt > 198:
-                sub_arr = arr[a:b, c:d]
-                cm = np.nanmedian(sub_arr)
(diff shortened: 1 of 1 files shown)


gmloose:
> The efficiency of bstat should be optimized instead. (I will extract few cases and see how well `bstat` actually works, but thats another issue).

Please note that there is a C++ implementation for `bstat` also. The `numpy` version was favored (what I read from the comments) because it out-performed the C++ version. But maybe there's room to improve the C++ version. OTOH, I wouldn't mind if we could turn PyBDSF into a pure-python package in the future. Especially the Fortran code could become a nightmare to maintain.

gmloose:
> That PR was a regression. It ignored the clipped statistics from `bstat` and uses the unclipped statistics, which should actually be worse, despite MAD is used now. 

It happens. No problem. Maybe I should have paid more attention.

gmloose (approved):

AlexKurek:
As for bstat, I think it could be replaced by:
astropy.stats.sigma_clipped_stats
Parameters:
cenfunc='median', stdfunc='mad_std', maxiters=3
or similar.

PR #276, opened 2026-05-14
Make `check_islands_for_overlap` faster

AlexKurek:
It makes the entire run faster by ~2x. I think the greater the size of the image, the greater the speed increase.

For 1.5 GB FITS,  19845 x 19845 pixels:
Out of 25,889 lines in the .gaul file, there are 4 that have numerical differences at the 6th decimal place.

before:
```
PyBDSF running time: 0:41:0
PyBDSF running time: 0:38:12
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        4 1005.956  251.489 1019.746  254.936 /home/akurek/miniconda3/envs/pybdsf-313_local_temp/lib/python3.13/site-packages/bdsf/wavelet_[atrous.py:681](http://atrous.py:681/)(check_islands_for_overlap)
```
after:
```
PyBDSF running time: 0:22:28
PyBDSF running time: 0:20:54
        4    3.252    0.813   15.651    3.913 /home/akurek/miniconda3/envs/pybdsf-313_local_temp/lib/python3.13/site-packages/bdsf/wavelet_[atrous.py:681](http://atrous.py:681/)(check_islands_for_overlap)
        4    3.020    0.755   15.355    3.839 /home/akurek/miniconda3/envs/pybdsf-313_local_temp/lib/python3.13/site-packages/bdsf/wavelet_[atrous.py:681](http://atrous.py:681/)(check_islands_for_overlap)
```

PyBDSF options used: https://github.com/user-attachments/files/27450974/run_pybdsf.py.

PR #440, opened 2026-09-12
Fix regression in `check_islands_for_overlap`

AlexKurek:
https://github.com/lofar-astron/PyBDSF/pull/276 made PyBDSF ~2x faster, so I would not revert it. But updating masks after island merges was omitted. As a result, when an existing island expanded due to a merge, `img.pyrank` and mask `orig_rankim_bool` retained stale state.
Subsequent islands overlapping these newly expanded boundaries read False from the stale mask, failing the overlap check and creating duplicate islands.


darafferty:
Ah, right, makes sense. I guess it's a fairly rare situation so easy to miss (at least the islands don't change for my test image).

darafferty (approved):

PR #441, opened 2026-09-13
Fix off by one in island lookup for wavelet Gaussians

AlexKurek:
**Bug**
When `atrous_orig_isl=True`, the algorithm looks up which island contains a wavelet Gaussian: `img.pyrank[int(g.centre_pix[0] + 1), int(g.centre_pix[1] + 1)]`.
Because `g.centre_pix` is already a 0-indexed coordinate matching `img.pyrank`, adding 1 shifts the position by 1 pixel in both dimensions.

**Fix**
Removed the +1 offset, used `round()` to find the nearest pixel and added non-negative boundary checks to prevent wrap-around.


gmloose (approved):
LGTM.

PR #442, opened 2026-09-13
Fix mask in FDR vs Hard threshold selection

AlexKurek:
`false_p` uses only the unmasked pixel count (`size`) while `source_p` (derived from the total image area, `img.omega`) does not account for the mask. This inflates `source_p` relative to `false_p`, biasing the algorithm toward Hard thresholding as the masked area grows.

**Fix**
Scale `source_p` proportionally to the fraction of unmasked pixels (`size / total_size`) before comparing it to `false_p`.

This could not cause https://github.com/lofar-astron/PyBDSF/issues/258.


gmloose (commented):
Makes sense. @darafferty, do you agree?

darafferty:
Yes, agreed, looks good. 

On a related note, maybe we should consider altering out test image to include some masked regions, as I don't think we're testing the masking code currently in CI?

PR #444, opened 2026-09-17
Add TODO comments

AlexKurek:
This is just to mark some spots where in the future improvements could be implemented.


gmloose:
I think it would be good to actually start these comments with `TODO:`. There are several tools that trigger on that specific string. And it also make it much easier to just `grep` for things that are still to-do.

PR #445, opened 2026-09-17
Fix `img.blankpix` calculation

AlexKurek:
The legacy assignment `img.blankpix = N.sum(mask)` was left in the code. Since `mask` can be set to `img.rms_mask`, it incorrectly included bright astronomical sources in the blank pixel count.

If there were no pixels outside the FOV (`noutside_univ == 0`), `img.blankpix` was never initialized, causing a crash during `if img.blankpix == 0:`.


bdsf/preprocess.py
```diff
@@ -101,8 +101,10 @@ def __call__(self, img):
 
         ### max/min pixel value & coordinates
         shape = image.shape[0:2]
-        if mask is not None:
-            img.blankpix = N.sum(mask)
+        if hasattr(img, 'mask_arr') and img.mask_arr is not None:
+            img.blankpix = N.sum(img.mask_arr)
+        else:
+            img.blankpix = 0
```
gmloose, line 105: I may be overlooking something here, but this patch feels overly complex. I think we already have all the required information at hand. AFAICS, the assignment of `img.blankpix` can be done around line 69, as `else`-clause of the `if noutside_univ > 0:` statement in line 54. Something like:
```
        if noutside_univ > 0:
            :
            :
            img.blankpix = N.sum(img.mask_arr)
        else:
            img.blankpix = 0
```

AlexKurek: If an image has noutside_univ == 0 but still contains NaN or blanked pixels (which are stored in img.mask_arr), setting img.blankpix = 0 in the else branch would ignore those blank pixels.
How about simplifying it to:
```
if getattr(img, 'mask_arr', None) is not None:
    img.blankpix = N.sum(img.mask_arr)
else:
    img.blankpix = 0
```
This keeps img.blankpix independent of the FOV calculation.

gmloose: OK, fair enough. Do you then also intend to remove line 68?

AlexKurek: Good catch, its not needed any more.

gmloose (approved):
LGTM.

PR #446, opened 2026-09-17
Fix flux and RMS calculations in plotting

AlexKurek:
Panning outside the image produced negative indices (e.g., `xmax = -2`). NumPy interpreted these as offsets from the end of the array, slicing almost the entire image and corrupting flux/RMS calculations.


gmloose (approved):
LGTM.

PR #447, opened 2026-09-18
Remove from __future__ import statements

gmloose:
Removed all `from __future__ import ...` statements. These were necessary in the transition from Python 2 to Python 3, which is now already more than six years behind us.


darafferty (approved):
Looks good.

PR #448, opened 2026-09-19
Fix initial Gaussian guess on the peak

AlexKurek:
The initial guess for the Gaussian fit places the centre at the middle of the fitting window. This is only correct when the window is centred on the peak. When the window is clipped by the island bounding box (`blc` clamped to 0, `trc` clamped to `n - 1` / `m - 1`), the peak is no longer at the window centre.

**Fix**
Use the peak position in window coordinates instead.
For an unclipped window `maxx - blc[0] == (s_imsize[0]-1)/2`, so results are identical there. Only sources whose window is clipped by the island edge are affected.


gmloose (commented):
I think this is OK. But I was wondering: `p_ini` is also used in two other calls to `fit_gaus2d` : https://github.com/lofar-astron/PyBDSF/blob/a4e6c40757257b83d16ee0060c0b13ee2c48e5ad/bdsf/gausfit.py#L693 and https://github.com/lofar-astron/PyBDSF/blob/a4e6c40757257b83d16ee0060c0b13ee2c48e5ad/bdsf/psf_vary.py#L235
Do these not suffer from this problem?

AlexKurek:
I think the other two are OK.

`gausfit.py:693` uses the peak location (`xf, yf` from the watershed markers) in the same frame as the fit - there's no window cropping there.

`psf_vary.py:235` fits `psfim`, each contributing source is shifted onto the array centre before stacking, and any source whose cutout would be clipped by the image edge is dropped.

PR #449, opened 2026-09-19
Fix ZeroDivisionError in moment analysis

AlexKurek:
The original implementation computed the 4-quadrant position angle manually:

1. `atan(abs(y / x))` calculated a reference angle in $[0, \pi/2]$.
2. `atanproper` mapped this angle into $[0, 2\pi)$ using conditional checks based on the signs of $x$ (`m2[0] - m2[1]`) and $y$ (`2.0 * m11`).

Replacing this with `np.arctan2(y, x) % (2 * pi)` is mathematically identical:

* `np.arctan2(y, x)` computes the 4-quadrant angle in $[-\pi, \pi]$.
* The `% (2 * pi)` modulo converts negative angles to $[0, 2\pi)$, matching the mapping of `atanproper`.

It prevents crashes when $x = 0$ (`m2[0] == m2[1]`, as `np.arctan2` handles zero denominators.


gmloose (approved):
I think it's safe to remove `atanproper` from `bdsf/functions.py`. I don't think it's being used outside PyBDSF, and this was the only place in this repository where it was used.

AlexKurek:
It is also used in `gdist_pa`.

PR #450, opened 2026-09-19
Fix Gaussian deconvolution error handling in `deconv2`

AlexKurek:
Unphysical Gaussian deconvolutions are incorrectly classified as point-like sources (`ifail = 1`) instead of failed/unphysical deconvolutions (`ifail = 2`).

In `if alpha < 0 or beta < 0 or s < t` the condition $s < t$ is guaranteed, meaning $s - t < 0$.

```
if 0.5 * (s - t) < limit and alpha > -limit and beta > -limit:
    ifail = 1
```
compares the negative minor eigenvalue `0.5 * (s - t)` against positive `limit` ($> 0$).

`0.5 * (s - t) < limit` is `True` for all inputs. The check fails to filter out unphysical deconvolutions.

Side note: here the author states, that this implementation was not working well: https://pwkit.readthedocs.io/en/stable/_modules/pwkit/astutil/.


bdsf/functions.py
```diff
@@ -522,11 +522,16 @@ def deconv2(gaus_bm, gaus_c):
             bmaj = sqrt(0.5*(s+t))
             bpa = rad * 0.5 * atan2(-gamma, alpha-beta)
         bmin = 0.0
-        if 0.5*(s-t) < limit and alpha > -limit and beta > -limit:
+        # https://github.com/lofar-astron/PyBDSF/pull/450
+        if 0.5*(s-t) < -limit and alpha > -limit and beta > -limit:
```
gmloose: Are you sure this is the correct fix. Shouldn't this be:
```
if 0.5 * (s - t) > -limit and alpha > -limit and beta > -limit:
```

AlexKurek: Yes, that`s right, thank you.

gmloose (approved):

